PR #36, opened 2015-02-25
fix(server.po): recover email string translations

zaach:
@shane-tomlinson @mathjazz I've added the translations back and updated the template.

### DIFF
--- a/locale/af/LC_MESSAGES/server.po
+++ b/locale/af/LC_MESSAGES/server.po
@@ -2,7 +2,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2015-02-20 19:15+0000\n"
+"POT-Creation-Date: 2015-02-25 18:58+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
 "Language: \n"
@@ -15,7 +15,8 @@ msgstr ""
 #: server/templates/pages/src/503.html:1
 #: server/templates/pages/src/index.html:1
 #: server/templates/pages/src/privacy.html:1
-#: server/templates/pages/src/terms.html:1
+#: server/templates/pages/src/terms.html:1 templates/reset.html:1
+#: templates/unlock.html:1 templates/verify.html:1
 msgid "Firefox Accounts"
 msgstr ""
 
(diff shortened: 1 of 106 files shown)


mathjazz:
Looks good. Just a quick question. The following pairs are almost identical:

Reset password - Reset password:
Verify account - Verify account:
Verify - Verify:

Are they intentionally different or can we unify them?

shane-tomlinson:
> Looks good. Just a quick question. The following pairs are almost identical:
>
> Reset password - Reset password:
> Verify account - Verify account:
> Verify - Verify:
> 
> Are they intentionally different or can we unify them?

@ryanfeeley - what do you think about some unification? "Verify account" and "Verify" seem reasonable.

@mathjazz - the colon is to separate the text from a verification link that follows. What is the best practice there?

mathjazz:
OK, in this case "string - string:" pairs should be separate strings, as they could be different parts of speech (verb vs. noun).

shane-tomlinson:
LGTM! @jrgm are these the strings that will be used for today's prod push?

zaach:
The prod push is using an earlier hash– this will just affect deployments using fxa-dev.

mathjazz:
That means some strings will be left untranslated after the new release or is this already the case in prod?

zaach:
@mathjazz We're using this hash: 9b283e26aaead44ef6c4d1216b0a6d01f547399f, which has the email strings but won't have translations to strings made in the last 5 days. But that's usually the case.

mathjazz:
OK. To make "that's usually the case" clear - we usually push to production first and give strings to localizers after that?

zaach:
@mathjazz Yeah, actually I merge strings for train 32 earlier than necessary. We could wait and merge them after the current train is pushed to production. As it stands, localizers will have extra time for train 32.

mathjazz:
Okay, thanks for explaining this @zaach!

jrgm:
Slightly off-topic, but about L10N and production releases, I wrote this micro-note on how the rpm is built with a specific version of fxa-content-server-l10n (default is HEAD). https://github.com/mozilla-services/svcops/blob/master/services/firefox-accounts/fxa-content-server/README.md

shane-tomlinson:
@zaach - it looks like this rebased or done again. A few commits came in after this PR was created and a clean merge is no longer possible.

zaach:
Rebased.

shane-tomlinson:
Thanks @zaach, @mathjazz, what do you think, can we hit "merge"?

mathjazz:
Let's hit the merge!

zaach:
Merging :+1:

mathjazz:
Verbatim updated.

But -- it seems like some locales aren't in sync with the templates:
https://localize.mozilla.org/projects/accounts/

(Sort by "Total".)

zaach:
@mathjazz Thanks, I've opened another PR. Not sure how those didn't make it into the initial train 32 PR.

PR #41, opened 2015-03-11
merge strings for train 33, with email subjects

shane-tomlinson:
@mathjazz - @jrgm realized when we switched repos where the email strings are extracted from, we failed to extract the email titles. Here is a PR targeting train-33 which re-adds the email titles. 

Many locales had all but one title already translated. The new title is "Re-verify your account".

Sorry for the second PR of the cycle. Hopefully email templates are all sorted now.


mathjazz:
OK, go ahead with the merge!

shane-tomlinson:
@mathjazz - merged!

mathjazz:
Verbatim updated.

PR #43, opened 2015-03-19
merge strings for train 34

shane-tomlinson:
@mathjazz - Strings for train 34 are extracted!

New additions are the account locked strings, and it looks like emails were properly extracted.


mathjazz:
Go! :)

PR #50, opened 2015-05-26
Strings extraction script

eoger:
Related to https://github.com/mozilla/fxa-content-server/issues/2398

### DIFF
--- a/scripts/extract_strings.sh
+++ b/scripts/extract_strings.sh
@@ -1,0 +1,87 @@
+#!/bin/sh
+
+# syntax:
+# extract_strings.sh [--mailer-repo ./fxa-auth-mailer] [--content-repo ./fxa-content-server] [--l10n-repo ./fxa-content-server-l10n] train_number
+
+set -e
+
+function usage() {
+    echo "syntax:"
+    echo "extract_strings.sh [--mailer-repo ./fxa-auth-mailer] [--content-repo ./fxa-content-server] [--l10n-repo ./fxa-content-server-l10n] train_number"
+    exit 1
+}
+
+function check_folder() {
+    if [[ ! -d $1 ]]; then
+        echo "Error: No such directory"
+        exit 1
+    else
+        echo "Ok!"
+    fi
+}
+
+if [[ $# -lt 1 ]]; then
+    usage;
+fi
+
+MAILER_DIR="./fxa-auth-mailer"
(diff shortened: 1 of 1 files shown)


scripts/extract_strings.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/sh
+
+# syntax:
+# extract_strings.sh [--mailer-repo ./fxa-auth-mailer] [--content-repo ./fxa-content-server] [--l10n-repo ./fxa-content-server-l10n] train_number
+
```
jrgm: May I suggest `set -o errexit` (a.k.a., `set -e`)

scripts/extract_strings.sh
```diff
@@ -0,0 +1,78 @@
...
+(cd $CONTENT_DIR && grunt l10n-extract) &&
+cp -r $CONTENT_DIR/locale/templates/ $L10N_DIR/locale/templates &&
+
+cd $L10N_DIR &&
+git checkout -b merge-train-$TRAIN_NUMBER-strings &&
+./scripts/merge_po.sh ./locale &&
+git add . &&
+git commit -m "merge strings for train $TRAIN_NUMBER" &&
+git push origin merge-train-$TRAIN_NUMBER-strings
```
vladikoff: I suggest not pushing automatically. Instead print the command that the developer should use. 
This allows us to make sure everything went okay during the extraction process before pushing.

In addition I don't use `origin` in some cases, but instead I use `upstream`

scripts/extract_strings.sh
```diff
@@ -0,0 +1,78 @@
...
+
+function usage() {
+    echo "syntax:"
+    echo "extract_strings.sh [--mailer-repo ./fxa-auth-mailer] [--content-repo ./fxa-content-server] [--l10n-repo ./fxa-content-server-l10n] train_number"
+    exit 1
+}
+
+function check_folder() {
+    if [[ ! -d $1 ]]; then
+        echo "Error!"
```
vladikoff: nit: add a more descriptive error here. `Error: failed to find a directory to extract from` or something

eoger:
Updated, thanks guys.

zaach:
Awesome! Tested it out today and it worked well.

eoger:
:+1:

shane-tomlinson:
This makes me happy.

PR #62, opened 2015-07-27
feat(client): scan emails for invalid style attr

TDA:
Add grunt to l10n repo
Add a grunt task to scan for invalid style attributes
Fixes #67 in auth-mailer repo
https://github.com/mozilla/fxa-auth-mailer/issues/67
@vladikoff 
@pdehaan

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+node_modules
+.tmp
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,0 +1,18 @@
+#!/usr/bin/env node
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = function (grunt) {
+  // load all grunt tasks based on environment
+  // show elapsed time at the end
+  require('time-grunt')(grunt);
+  require('load-grunt-tasks')(grunt);
+
+  grunt.initConfig({
+    pkg: grunt.file.readJSON('package.json')
+  });
+
+  grunt.loadTasks('grunttasks');
+};
--- a/grunttasks/po2json.js
+++ b/grunttasks/po2json.js
@@ -1,0 +1,74 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// grunt task to create .json files out of .po files.
+// .po files are expected to already be downloaded.
+
+var path = require('path');
+var i18n = require('i18n-abide');
+
+module.exports = function (grunt) {
+  grunt.config('po2json', {
+    options: {
+      format: 'raw',
+      fuzzy: true,
+      output_filename: function (file) { //eslint-disable-line camelcase
+
+        /**
+         * the files are stored in the locale subdirectory with a directory
+         * structure of:
+         * locale/
+         *       <locale_name>/
+         *                    LC_MESSAGES/
+         *                               server.po
+         *                               client.po
+         *
+         * Each locale is stored in its own subdirectory in the output i18n
+         * directory.
+         **/
+        var matches = /^locale\/([^\/]+)\/LC_MESSAGES\/(.*)$/.exec(file);
+
+        // In order to make sure the locale is in a form that i18n-abide expects
+        // we convert it from locale to langauge back to locale.
+        var locale = i18n.normalizeLocale(matches[1]);
+        var filename = matches[2];
+        if (filename === 'server.po') {
+          filename = 'messages.json';
+        } else {
+          // get rid of the .po extension, replace with .json
+          filename = path.basename(filename, '.po') + '.json';
+        }
+        return locale + '/' + filename;
+      },
+      output_transform: function (data) { //eslint-disable-line camelcase
+        // write the first translation only, ignore pluralization.
+        var isArray = function (item) {
+          return Object.prototype.toString.call(item) === '[object Array]';
+        };
+
+        var transformed = {};
+        for (var msgid in data) {
+          var translation = data[msgid];
+          if (isArray(translation) && translation.length >= 2) {
+            translation = translation[1];
+          }
+
+          transformed[msgid] = translation;
+        }
+
+        return transformed;
+      }
+    },
+    all: {
+      src: ['locale/**/**/*.po'],
+      dest: '.tmp/l10n/'
+    },
+    template: {
+      src: ['locale/templates/**/*.pot'],
+      dest: '.tmp/l10n'
+    }
+  });
+};
+
+
--- a/grunttasks/scan-email-translations.js
+++ b/grunttasks/scan-email-translations.js
@@ -1,0 +1,115 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = function (grunt) {
+  grunt.registerTask('scan-email-translations', 'Scan the generated emails for invalid style tags', [
+    'generate-valid-styles',
+    'po2json',
+    'verify-styles'
+    ]
+  );
+
+  grunt.registerMultiTask('verify-styles', 'Verify whether the translated styles are valid', function (){
+    var options = this.options({
+      validStyles: []
+    });
+    var styleArray = options.validStyles;
+
+    var fileArray = this.files[0].src;
+
+    var hasErrors = 0;
+    var styleRegex = /style="[\s\S]*?"/ig;
+
+    fileArray.forEach(function (src) {
+      var json = grunt.file.readJSON(src);
+      var styleList = [];
+      for(var key in json) {
+        if(key.indexOf('style') > 0) {
+          styleList.push(key);
+        }
+      }
+      // Check if there were any styles found, .match() might return null
+      if (styleList !== null) {
+        styleList = styleList.map(stripSpecialChars);
+        // retrieve the style attribute and its contents and store
+        styleList = styleList.map(function (str) {
+          if (styleRegex.test(str)) {
+            str = str.match(styleRegex)[0];
+            return str;
+          }
+        });
+        styleList.forEach(function (style) {
+          if(style !== undefined){
+            if (styleArray.indexOf(style) === -1) {
+              grunt.log.error('Found a style which should not be there', style, src);
+              hasErrors = hasErrors + 1;
+            }
+          }
+        });
+      }
+    });
+    if (hasErrors !== 0) {
+      grunt.fail.warn('Found ' + hasErrors + ' ' + grunt.util.pluralize(hasErrors, 'error/errors') + ' in ' + fileArray.length + ' files');
+    } else {
+      grunt.log.writeln('Checked ' + fileArray.length + ' files for valid styles and found nothing wrong.');
+    }
+  });
+
+  grunt.config('verify-styles', {
+    dist: {
+      options: {
+        validStyles: grunt.file.readJSON('.tmp/validStyles.json')
+      },
+      files: [
+        {
+          expand: false,
+          src: [
+            '.tmp/l10n/*/*.json'
+          ]
+        }
+      ]
+    }
+  });
+
+
+  grunt.registerMultiTask('generate-valid-styles', 'Generate the list of valid styles for translation', function () {
+    //different regex from above, uses single quotes not double
+    var styleRegex = /style='[\s\S]*?'/ig;
+    var styleArray = [];
+
+    // iterate through .pot file and extract the styles
+    this.files[0].src.forEach(function (src) {
+      var contents = grunt.file.read(src);
+      if (styleRegex.test(contents)) {
+        Array.prototype.push.apply(styleArray, contents.match(styleRegex));
+      }
+    });
+    styleArray = styleArray.map(stripSpecialChars);
+    // write out the styles to a temporary JSON file
+    grunt.file.write(this.files[0].dest, JSON.stringify(styleArray));
+    grunt.log.writeln('Generated styles from ' + this.files[0].src.length + ' files');
+  });
+
+  grunt.config('generate-valid-styles', {
+    dist: {
+      files: [
+        {
+          expand: false,
+          src: [
+            'locale/templates/*/server.pot'
+          ],
+          dest: '.tmp/validStyles.json'
+        }
+      ]
+    }
+  });
+
+  function stripSpecialChars(str) {
+    str = str.replace(/'/g, '"');
+    str = str.replace(/""/g, '');
+    str = str.replace(/\n/g, '');
+    return str.replace(/(\\)/g, '');
+  }
+
+};
--- a/package.json
+++ b/package.json
@@ -13,5 +13,12 @@
   "license": "MPL-2.0",
   "bugs": {
     "url": "https://github.com/mozilla/fxa-content-server-l10n/issues"
+  },
+  "devDependencies": {
+    "grunt": "0.4.5",
+    "grunt-po2json": "git://github.com/shane-tomlinson/grunt-po2json.git#aa276503e",
+    "i18n-abide": "0.0.23",
+    "load-grunt-tasks": "3.2.0",
+    "time-grunt": "1.2.1"
   }
 }


Gruntfile.js
```diff
@@ -0,0 +1,22 @@
...
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = function (grunt) {
+  // load all grunt tasks based on environment
+  if (process.env.NODE_ENV && process.env.NODE_ENV === 'production') {
+    require('load-grunt-tasks')(grunt, {scope: 'dependencies'});
+  } else {
```
pdehaan: Maybe just remove the `if..else` and replace w/:

```js
module.exports = function (grunt) {
  require('time-grunt')(grunt);
  require('load-grunt-tasks')(grunt);

  // ...
};
```

... And I'm not even sure if we need either of those since we aren't loading any external Grunt tasks.

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,85 @@
...
+
+  grunt.config('generate-valid-styles', {
+    dist: {
+      files: [
+        {
+          expand: false,
+          src: [
+            'locale/templates/*/server.pot'
+          ],
+          dest: '.tmp/validStyles.json'
```
pdehaan: Something tells me, we'll want to add .tmp/ to a .gitignore file.

TDA:
@pdehaan Done, and done!

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,85 @@
...
+    ]
+  );
+
+  grunt.registerMultiTask('verify-styles', 'Verify whether the translated styles are valid', function (){
+    var fileArray = this.files[0].src;
+    var contents = grunt.file.read(fileArray.shift());
+    var styleArray = JSON.parse(contents);
+
+    var hasErrors = 0;
+    var styleRegex = /style='(.*)'/ig;
```
pdehaan: Not sure this RegExp will work. I think it's too greedy.
I tried reversing the order of the attributes in the `<a>` tag and my .tmp/validStyles.json file looks like:
```
["style='color: #0095dd; text-decoration: none;' href='https://support.mozilla.org'"]
```

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,85 @@
...
+            'locale/*/*/server.po'
+          ]
+        }
+      ]
+    }
+  });
+
+
+  grunt.registerMultiTask('generate-valid-styles', 'Generate the list of valid styles for translation', function () {
+    var styleRegex = /style='(.*)'/ig;
```
pdehaan: See https://github.com/mozilla/fxa-content-server-l10n/pull/62/files#r35603540

Also, since this RegExp is duplicated, we may want to see if we can DRY it out.

grunttasks/scan-email-translations.js
rfk: Do we specifically need to check the style attr, or can we just pull out the entire tag and match on that?  For example, if the localization contains `"<a style='foo bar' href='whatever'>" then can we just assert that this entire tag must appear in the original template?  This may be less susceptible to edge-cases.

TDA:
@pdehaan If you reverse the attr and it matches them in the .pot file, it will also match it in the .po file, no? I could add a '>' at the end to make it less greedy, but changing the order of attributes would always make it capture additional text, unless I put in a ';' somewhere there and restrict it to styles. Problems of HTML not being a regular language :) And I really dont think adding a HTML parser just for one tag is worth it.

grunttasks/scan-email-translations.js
pdehaan: @TDA I think we can make it non-greedy by changing the RegExp to something like this:
```js
var styleRegex = /style='(.*?)'/ig;
```

That way it will stop on the first `'` it encounters. But yeah, maybe we need to rethink this and do what @rfk suggests, which is comparing the entire HTML tag's contents. Because I can break the above RegExp by using double quotes instead of single quotes, and turns out there's an even bigger edge case of multiline tags.

For example, I was curious why I could completely edit the `style` attribute in the French locale and it wouldn't detect it, and then after adding a few `console.log()` in the Grunt files, I think it is because the `style=''` starts on one line and closes on another line (which seems to explain why in some cases `styleList` had 1 result and in other cases had 2 -- where the 1 result was probably where it matches the English key in the .po files):

```pot
#: templates/verify.html:7
msgid ""
"For more information, please visit <a href='https://support.mozilla.org' "
"style='color: #0095dd; text-decoration: none;' target='nein'>Mozilla Support</a>"
msgstr ""
"Pour davantage d'informations, veuillez consulter <a "
"href='https://support.mozilla.org' style='color: green; display: none; text-decoration: "
"none;'>l'assistance de Mozilla</a>"
```

We'd have to some more research, but we may need to do something like a [**grunt-po2json**](https://github.com/rockykitamura/grunt-po2json) where we convert the .po files to .json and then do the scanning.

.gitignore
```diff
@@ -0,0 +1 @@
+.tmp
```
vladikoff: add `node_modules` to gitignore as well

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,85 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = function (grunt) {
+  grunt.registerTask('scan-email-translations', 'Scan the generated emails for invalid style tags', [
```
vladikoff: this only does "emails"? what about other stuff that we need to move from https://github.com/mozilla/fxa-content-server/pull/2742/files ?

grunttasks/scan-email-translations.js
vladikoff: > For example, I was curious why I could completely edit the style attribute in the French locale and it wouldn't detect it, 

Just to confirm, we must match on everything in the  style attribute, i.e the name "style" and all of its content.

grunttasks/scan-email-translations.js
TDA: @vladikoff This does only the emails, the other ones will be in a separate PR for this repo. I am working on those.

.gitignore
TDA: added!

TDA:
@pdehaan changes made as suggested! :) r? @vladikoff

grunttasks/po2json.js
```diff
@@ -0,0 +1,74 @@
...
+        return transformed;
+      }
+    },
+    all: {
+      src: ['locale/**/**/*.po'],
+      dest: '.tmp/l10n/'
+    },
+    template: {
+      src: ['locale/templates/**/*.pot'],
+      dest: '.tmp/i18n'
```
pdehaan: weird, wonder if this was in the other repo too, but this seems odd that one goes into `l10n/**` and the other goes into `i18n/**`.

TDA: No, that was me, to prevent the overwriting, i can change it back :octopus:

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,108 @@
...
+  });
+
+  grunt.config('verify-styles', {
+    dist: {
+      files: [
+        {
+          expand: false,
+          src: [
+            '.tmp/validStyles.json',
+            '.tmp/l10n/*/messages.json'
```
pdehaan: Do we want to lint the `.tmp/l10n/*/client.json` files as well, or just `messages.json`?
Also, do we need to verify the validStyles.json file since that was already generated via `generate-valid-styles` task?

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,108 @@
...
+  grunt.registerMultiTask('verify-styles', 'Verify whether the translated styles are valid', function (){
+    var fileArray = this.files[0].src;
+    var contents = grunt.file.read(fileArray.shift());
+    var styleArray = JSON.parse(contents);
+
+    var hasErrors = 0;
+    var styleRegex = /style="(.*?)"/ig;
+
+    fileArray.forEach(function (src) {
+      var json = JSON.parse(grunt.file.read(src));
```
pdehaan: nit: `var json = grunt.file.readJSON(src);`

TDA: changed! @pdehaan I knew there must be an easier way :D

grunttasks/scan-email-translations.js
TDA: client.json has no style attributes? I would gladly add it in :D

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,108 @@
...
+    'generate-valid-styles',
+    'po2json',
+    'verify-styles'
+    ]
+  );
+
+  grunt.registerMultiTask('verify-styles', 'Verify whether the translated styles are valid', function (){
+    var fileArray = this.files[0].src;
+    var contents = grunt.file.read(fileArray.shift());
+    var styleArray = JSON.parse(contents);
```
TDA: @pdehaan Do you want me to combine these two into one with readJSON?

pdehaan: Sure, you could (assuming it works).
I have to admit, I am/was a bit lost as to what this `shift()` magic is. I think we're adding too much sorcery to the `src` input if we're trying to pluck the first parameter as the config file to lint against.

pdehaan: Wildly untested counter-proposal. Instead of assuming the first item in the `src` array is your `validStyles` file/array, just pass the valid styles in as an option:

```js
  grunt.config('verify-styles', {
    dist: {
      options: {
        validStyles: grunt.file.readJSON('.tmp/validStyles.json')
      },
      files: [
        {
          expand: false,
          src: [
            '.tmp/l10n/*/messages.json'
          ]
        }
      ]
    }

  ...

  grunt.registerMultiTask('verify-styles', 'Verify whether the translated styles are valid', function (){
    var fileArray = this.files[0].src;
    var options = this.options({
      validStyles: []
    });
    var styleArray = options.validStyles;

    var hasErrors = 0;
    var styleRegex = /style="(.*?)"/ig;
```

shane-tomlinson:
When I run either of the new tasks, I receive an error:

```bash
➜   grunt scan-email-translations
Loading "scan-email-translations.js" tasks...ERROR
>> Error: Unable to read ".tmp/validStyles.json" file (Error code: ENOENT).
Warning: Task "generate-valid-styles" not found. Use --force to continue.

Aborted due to warnings.


Execution Time (2015-08-06 13:19:06 UTC)
loading tasks  257ms  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 100%
Total 257ms

stomlinson@shane-office: /Users/stomlinson/development/fxa-content-server-l10n git:(invalid-style)
➜   grunt verify-styles
Loading "scan-email-translations.js" tasks...ERROR
>> Error: Unable to read ".tmp/validStyles.json" file (Error code: ENOENT).
>> No "verify-styles" targets found.
Warning: Task "verify-styles" failed. Use --force to continue.

Aborted due to warnings.


Execution Time (2015-08-06 13:19:12 UTC)
loading tasks  257ms  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 100%
Total 257ms

```

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,115 @@
...
+        Array.prototype.push.apply(styleArray, contents.match(styleRegex));
+      }
+    });
+    styleArray = styleArray.map(stripSpecialChars);
+    // write out the styles to a temporary JSON file
+    grunt.file.write(this.files[0].dest, JSON.stringify(styleArray));
+    grunt.log.writeln('Generated styles from ' + this.files[0].src.length + ' files');
+  });
+
+  grunt.config('generate-valid-styles', {
```
shane-tomlinson: All of these tasks in one file makes them difficult to find. Can you split this file?

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,115 @@
...
+          ]
+        }
+      ]
+    }
+  });
+
+
+  grunt.registerMultiTask('generate-valid-styles', 'Generate the list of valid styles for translation', function () {
+    //different regex from above, uses single quotes not double
+    var styleRegex = /style='[\s\S]*?'/ig;
```
shane-tomlinson: What happens if we somehow forget to use a `'` and instead use a `"`, this regexp will not catch that?

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,115 @@
...
+    var styleArray = [];
+
+    // iterate through .pot file and extract the styles
+    this.files[0].src.forEach(function (src) {
+      var contents = grunt.file.read(src);
+      if (styleRegex.test(contents)) {
+        Array.prototype.push.apply(styleArray, contents.match(styleRegex));
```
shane-tomlinson: This seems like a very round about way of doing:

```js
styleArray = styleArray.concat(contents.match(styleRegex));
```

Which I think can be made more clear anyways:

```js
var matches = contents.match(styleRegex);
if (matches) {
  styleArray = styleArray.concat(matches);
}

File: grunttasks/scan-email-translations.js [hunk 14]
```diff
@@ -0,0 +1,115 @@
...
+      }
+    });
+    styleArray = styleArray.map(stripSpecialChars);
```
shane-tomlinson: Why are you stripping the special chars? The reason is not obvious.

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,115 @@
...
+            '.tmp/l10n/*/*.json'
+          ]
+        }
+      ]
+    }
+  });
+
+
+  grunt.registerMultiTask('generate-valid-styles', 'Generate the list of valid styles for translation', function () {
+    //different regex from above, uses single quotes not double
```
shane-tomlinson: How is the list of valid styles used elsewhere, that is not obvious.

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,115 @@
...
+    grunt.log.writeln('Generated styles from ' + this.files[0].src.length + ' files');
+  });
+
+  grunt.config('generate-valid-styles', {
+    dist: {
+      files: [
+        {
+          expand: false,
+          src: [
+            'locale/templates/*/server.pot'
```
shane-tomlinson: What about `client.pot`?

shane-tomlinson: IIRC, `/*/` will only traverse down a single directory, which makes an assumption that the directory structure will remain the same. I think this should be changed to `/**/`

grunttasks/scan-email-translations.js
shane-tomlinson: This string is duplicated in multiple places, make it into a constant.

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,115 @@
...
+  grunt.config('verify-styles', {
+    dist: {
+      options: {
+        validStyles: grunt.file.readJSON('.tmp/validStyles.json')
+      },
+      files: [
+        {
+          expand: false,
+          src: [
+            '.tmp/l10n/*/*.json'
```
shane-tomlinson: Same here about traversing down one directory, and the assumption of the directory structure remaining constant. `/*/` should probably be `/**/`.

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,115 @@
...
+      grunt.fail.warn('Found ' + hasErrors + ' ' + grunt.util.pluralize(hasErrors, 'error/errors') + ' in ' + fileArray.length + ' files');
+    } else {
+      grunt.log.writeln('Checked ' + fileArray.length + ' files for valid styles and found nothing wrong.');
+    }
+  });
+
+  grunt.config('verify-styles', {
+    dist: {
+      options: {
+        validStyles: grunt.file.readJSON('.tmp/validStyles.json')
```
shane-tomlinson: As in my case, if this file does not exist, this blows up during configuration.

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,115 @@
...
+  grunt.registerMultiTask('verify-styles', 'Verify whether the translated styles are valid', function (){
+    var options = this.options({
+      validStyles: []
+    });
+    var styleArray = options.validStyles;
+
+    var fileArray = this.files[0].src;
+
+    var hasErrors = 0;
+    var styleRegex = /style="[\s\S]*?"/ig;
```
shane-tomlinson: What happens if the translator used a `'` instead of a `"`, this regexp would let the style through unchecked.

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,115 @@
...
+
+    fileArray.forEach(function (src) {
+      var json = grunt.file.readJSON(src);
```
shane-tomlinson: What is in this JSON file? That is not documented anywhere.

shane-tomlinson: `json` is a pretty bad name that does not describe what is in the file. It only describes the type of the contents of the file.

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,115 @@
...
+      var styleList = [];
+      for(var key in json) {
+        if(key.indexOf('style') > 0) {
+          styleList.push(key);
+        }
+      }
+      // Check if there were any styles found, .match() might return null
+      if (styleList !== null) {
```
shane-tomlinson: `styleList` will *never* be null, you have already set it to `[]` above. This branch will always be taken.

What you probably want is:

```js
if (styleList.length) {
```

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,115 @@
...
+        styleList = styleList.map(stripSpecialChars);
+        // retrieve the style attribute and its contents and store
+        styleList = styleList.map(function (str) {
+          if (styleRegex.test(str)) {
+            str = str.match(styleRegex)[0];
```
shane-tomlinson: What happens if someone adds two `style` attributes to a string, only the first will be checked and the second will be allowed through w/o a check.

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,115 @@
...
+    var styleArray = options.validStyles;
+
+    var fileArray = this.files[0].src;
+
+    var hasErrors = 0;
+    var styleRegex = /style="[\s\S]*?"/ig;
+
+    fileArray.forEach(function (src) {
+      var json = grunt.file.readJSON(src);
+      var styleList = [];
```
shane-tomlinson: This styleList is from where, is this the list of allowed styles or the list of styles from the translations? Can you use a more specific name?

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,115 @@
...
+    'po2json',
+    'verify-styles'
+    ]
+  );
+
+  grunt.registerMultiTask('verify-styles', 'Verify whether the translated styles are valid', function (){
+    var options = this.options({
+      validStyles: []
+    });
+    var styleArray = options.validStyles;
```
shane-tomlinson: `styleArray` is a very generic name, I find `validStyles` that you use on the right much easier to understand.

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,115 @@
...
+
+    var fileArray = this.files[0].src;
```
shane-tomlinson: `fileArray` is very generic too, what about `filesToCheck` or similar.

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,115 @@
...
+        styleList = styleList.map(stripSpecialChars);
+        // retrieve the style attribute and its contents and store
+        styleList = styleList.map(function (str) {
+          if (styleRegex.test(str)) {
+            str = str.match(styleRegex)[0];
+            return str;
+          }
+        });
+        styleList.forEach(function (style) {
+          if(style !== undefined){
```
shane-tomlinson: Can `style` *ever* be undefined? Could it be null? Could it be a number? What happens here with non-string input?

shane-tomlinson: nit: add a space in `if(`

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,115 @@
...
+            'locale/templates/*/server.pot'
+          ],
+          dest: '.tmp/validStyles.json'
+        }
+      ]
+    }
+  });
+
+  function stripSpecialChars(str) {
+    str = str.replace(/'/g, '"');
```
shane-tomlinson: What is the deciding factor whether something is considered a special char?

grunttasks/scan-email-translations.js
```diff
@@ -0,0 +1,115 @@
...
+    str = str.replace(/""/g, '');
+    str = str.replace(/\n/g, '');
```
shane-tomlinson: what happens if someone uses `\r` `\t` or any other whitespace special character?

grunttasks/scan-email-translations.js
TDA: client.pot does not contain styles, do you want me to add it just in case? will correct directory stuff! @shane-tomlinson

grunttasks/scan-email-translations.js
TDA: That list is used to test the generated files in the valid-styles task.

grunttasks/scan-email-translations.js
TDA: If the styles were on separate lines, or separated by a different quote (single vs double) then they may escape validation, to prevent that, we need to strip off newlines and make the quotes uniform.
Tagging @pdehaan for finding the newline issue.

grunttasks/scan-email-translations.js
TDA: I added the Regex.test instead of String.match so that even if the contents were empty(or null or undefined), they wouldnt blow up. I can change the push to a concat though.

grunttasks/scan-email-translations.js
TDA: i will split the files :)

grunttasks/scan-email-translations.js
TDA: changed this to include both quotes.

grunttasks/scan-email-translations.js
TDA: On second thought,
` styleArray = styleArray.concat(contents.match(styleRegex)); `
actually appends each letter as an element in the array. So I am letting this as it is.

grunttasks/scan-email-translations.js
TDA: We are just normalizing the style strings to use double quotes and ignore newlines.

grunttasks/scan-email-translations.js
TDA: \t and \r added to the replace method.

grunttasks/scan-email-translations.js
TDA: done, and done :)

grunttasks/scan-email-translations.js
TDA: changed!

grunttasks/scan-email-translations.js
TDA: changed!

grunttasks/scan-email-translations.js
TDA: Excellent point!!! wow, I completely missed this. Added logic to concat the strings when multiple styles are found in a tag. @shane-tomlinson

grunttasks/scan-email-translations.js
TDA: changed!

grunttasks/scan-email-translations.js
TDA: changed!

grunttasks/scan-email-translations.js
TDA: changed to check for both quotes.

grunttasks/scan-email-translations.js
TDA: changed!

TDA:
I am killing this PR and moving the changes to the other PR here: https://github.com/mozilla/fxa-content-server-l10n/pull/63/
This is because the other one needs to be rebased anyway if the files are split, and all the xss-based changes will be in one single PR. 
Closing.

grunttasks/scan-email-translations.js
shane-tomlinson: > client.pot does not contain styles, do you want me to add it just in case? will correct directory stuff! @shane-tomlinson

Yeah, let's add it now since we are here, if it's not difficult.

PR #63, opened 2015-07-29
feat(client): sanitize html translations in l10n

### LINKED ISSUE #61: Move Grunt XSS tasks to the l10n repo
Per @vladikoff's https://github.com/mozilla/fxa-auth-mailer/issues/67#issuecomment-125216940

> _if the "full" templates live in the l10n repo then we should probably move those XSS parse tasks out of the content server and put them into the l10n repo_

TDA:
Ports the xss checking from the content-server repo to this repo.
Fixes #61 
@pdehaan 
@vladikoff

### DIFF
--- a/grunttasks/generate-valid-styles.js
+++ b/grunttasks/generate-valid-styles.js
@@ -1,0 +1,41 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var path = require('path');
+var stripSpecialChars = require('./lib/strip-special-chars');
+
+module.exports = function (grunt) {
+  'use strict';
+   grunt.registerMultiTask('generate-valid-styles', 'Generate the list of valid styles for translation', function () {
+    var validStylesRegex = /style=('|")[\s\S]*?('|")/ig;
+    var validStylesArray = [];
+
+    // iterate through .pot file and extract the styles
+    this.files[0].src.forEach(function (src) {
+      var contents = grunt.file.read(src);
+      if (validStylesRegex.test(contents)) {
+        validStylesArray = validStylesArray.concat(contents.match(validStylesRegex));
+      }
+    });
+    validStylesArray = validStylesArray.map(stripSpecialChars);
+    // write out the styles to a temporary JSON file
+    grunt.file.write(path.join(this.files[0].dest, grunt.config.process('<%= config.validStylesFile %>')), JSON.stringify(validStylesArray));
+    grunt.log.writeln('Generated styles from ' + this.files[0].src.length + ' files');
+  });
+
+  grunt.config('generate-valid-styles', {
+    dist: {
+      files: [
+        {
+          expand: false,
+          src: [
+            '<%= config.strings_src %>/templates/**/*.pot'
+          ],
+          dest: '<%= config.tmp %>'
+        }
+      ]
+    }
+  });
+
+};
--- a/grunttasks/json2html.js
+++ b/grunttasks/json2html.js
@@ -1,0 +1,59 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var path = require('path');
+
+// The xss module does not run on json, it needs to form a dom tree with the html tags internally so,
+// we generate those intermediate files.
+// Four steps are performed to lint the files:
+//  1. Clean the directories
+//  2. Generate the JSON files from the PO files
+//  3. The JSON files are then written out to temporary HTML files
+//  4. The HTML files are linted
+
+
+module.exports = function (grunt) {
+  'use strict';
+  grunt.registerMultiTask('l10n-json-to-html', 'Convert l10n JSON files to HTML', function () {
+    var files = 0;
+    this.files.forEach(function (file) {
+      var HTMLcontent = '';
+      var src = file.src[0];
+      var pathname = src.split(path.sep);
+      var localeStrings = grunt.file.readJSON(src);
+      for (var val in localeStrings) {
+        var value = localeStrings[val];
+        // check if the values will be translated, objects, numbers etc are not translated.
+        // so we dont have to write those out to the html file.
+        if (typeof value !== 'object' && value !== '' && isNaN(value)) {
+          HTMLcontent += value.toString() + '\n\n';
+        }
+      }
+      if (HTMLcontent !== '') {
+        grunt.file.write(path.join(grunt.config.process('<%= config.tmp %>'),
+                  grunt.config.process('<%= config.html %>'),
+                  pathname[1],
+                  pathname[2],
+                  pathname[3].replace('json', 'html')
+                  ),
+              HTMLcontent);
+        files = files + 1;
+      }
+    });
+    grunt.log.ok('%d %s converted from JSON to HTML', files, grunt.util.pluralize(files, 'file/files'));
+  });
+
+  grunt.config('l10n-json-to-html', {
+    dist: {
+      files: [
+        {
+          expand: true,
+          src: [
+            '<%= config.tmp %>/l10n/es/*.json'
+          ]
+        }
+      ]
+    }
+  });
+};
--- a/grunttasks/verify-styles.js
+++ b/grunttasks/verify-styles.js
@@ -1,0 +1,58 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var path = require('path');
+var stripSpecialChars = require('./lib/strip-special-chars');
+
+module.exports = function (grunt) {
+  'use strict';
+  grunt.registerMultiTask('verify-styles', 'Verify whether the translated styles are valid', function (){
+    var validStylesFile = path.join(this.files[0].dest, grunt.config.process('<%= config.validStylesFile %>'));
+    var validStylesArray = grunt.file.readJSON(validStylesFile);
+    var filesToBeChecked = this.files[0].src;
+
+    var hasErrors = 0;
+    var generatedStylesRegex = /style=('|")[\S\s]*?('|")/ig;
+
+    filesToBeChecked.forEach(function (src) {
+      var HTMLContent = grunt.file.read(src);
+      var generatedStylesList = [];
+      if (generatedStylesRegex.test(HTMLContent)) {
+        generatedStylesList = generatedStylesList.concat(HTMLContent.match(generatedStylesRegex));
+      }
+      // Check if there were any styles found, .match() might return null
+      if (generatedStylesList.length > 0) {
+        generatedStylesList = generatedStylesList.map(stripSpecialChars);
+        // retrieve the style attribute and its contents and store
+      }
+      generatedStylesList.forEach(function (style) {
+          if(style && style.length > 0){
+            if (validStylesArray.indexOf(style) === -1) {
+              grunt.log.error('Found a style which should not be there: %s\n File: %s\n', style, src);
+              hasErrors = hasErrors + 1;
+            }
+          }
+        });
+    });
+    if (hasErrors !== 0) {
+      grunt.fail.warn('Found ' + hasErrors + ' ' + grunt.util.pluralize(hasErrors, 'error/errors') + ' in ' + filesToBeChecked.length + ' files');
+    } else {
+      grunt.log.writeln('Checked ' + filesToBeChecked.length + ' files for valid styles and found nothing wrong.');
+    }
+  });
+
+    grunt.config('verify-styles', {
+    dist: {
+      files: [
+        {
+          expand: false,
+          src: [
+            '<%= config.tmp %>/<%= config.html %>/l10n/**/*.html'
+          ],
+          dest: '<%= config.tmp %>'
+        }
+      ]
+    }
+  });
+};
--- a/grunttasks/xss-mitigate.js
+++ b/grunttasks/xss-mitigate.js
@@ -1,0 +1,76 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var xss = require('xss');
+
+module.exports = function (grunt) {
+  'use strict';
+  grunt.registerTask('xss-flag', 'Run the XSS parser on the generated l10n files', [
+    'po2json',
+    'l10n-json-to-html',
+    'xss-parse'
+  ]);
+
+  grunt.registerMultiTask('xss-parse', 'Parse HTML files to check for XSS vectors', function () {
+    var interpolatedValueRegex = /^\%\(\w+\)s$/;
+    this.files.forEach(function (file) {
+      // iterate through each HTML file in the .tmp folder
+      var src = file.src[0];
+      var contents = grunt.file.read(src);
+      // whitelist of allowed tags and attributes, nothing else is allowed
+      var options = {
+        whiteList: {
+          a: ['href', 'id', 'style'],
+          em: [],
+          span: ['id', 'tabindex'],
+          strong: []
+        },
+        onTagAttr: function (tag, name, value, isWhiteAttr) {
+          // On encountering any tag, check if the tag has a valid attribute.
+          // If the attribute is valid, check for its value, flag if illegal
+          // Illegal includes any mention of the string 'javascript', case insensitive
+          // Templated strings are allowed i.e strings of the form '%(termsURI)s'
+          // If these are considered to be not secure, we should change
+            if (name !== 'style' &&
+                isWhiteAttr &&
+                xss.safeAttrValue(tag, name, value) === '' &&
+                ! interpolatedValueRegex.test(value)) {
+              grunt.log.error('%s: INVALID VALUE FOR ATTRIBUTE <%s %s="%s">', src, tag, name, value);
+              return '';
+              // return '' to replace the invalid attribute with ''
+            }
+        },
+        onIgnoreTag: function (tag, html, options) {
+          // On encountering any illegal tag, flag it
+          grunt.log.error('%s: UNEXPECTED TAGS found <%s>', src, tag);
+          return '';
+        },
+        onIgnoreTagAttr: function (tag, name, value, isWhiteAttr) {
+          // On encountering any illegal attribute, flag it
+          grunt.log.error('%s: UNEXPECTED TAG ATTRIBUTES FOUND <%s %s="">', src, tag, name);
+          return '';
+        }
+      };
+      contents = xss(contents, options);
+    });
+    if (this.errorCount !== 0) {
+      grunt.fail.warn('Found ' + this.errorCount + ' ' + grunt.util.pluralize(this.errorCount, 'error/errors') + ' in ' + this.files.length + ' files');
+    } else {
+      grunt.log.writeln('Checked ' + this.files.length + ' files for XSS vulnerabilities');
+    }
+  });
+
+  grunt.config('xss-parse', {
+    dist: {
+      files: [
+        {
+          expand: true,
+          src: [
+            '<%= config.tmp %>/html/l10n/**/*.html'
+          ]
+        }
+      ]
+    }
+  });
+};
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+node_modules
+.tmp
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,0 +1,18 @@
+#!/usr/bin/env node
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = function (grunt) {
+  // load all grunt tasks based on environment
+  // show elapsed time at the end
+  require('time-grunt')(grunt);
+  require('load-grunt-tasks')(grunt);
+
+  grunt.initConfig({
+    pkg: grunt.file.readJSON('package.json')
+  });
+
+  grunt.loadTasks('grunttasks');
+};
--- a/grunttasks/config.js
+++ b/grunttasks/config.js
@@ -1,0 +1,17 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = function (grunt) {
+  var TEMP_DIR = '.tmp';
+  var HTML_DIR = 'html';
+  var VALID_STYLES_FILE_NAME = 'valid_styles.json';
+
+  grunt.config('config', {
+    /*eslint-disable camelcase */
+    html: HTML_DIR,
+    strings_src: 'locale',
+    tmp: TEMP_DIR,
+    validStylesFile: VALID_STYLES_FILE_NAME
+  });
+};
--- a/grunttasks/lib/strip-special-chars.js
+++ b/grunttasks/lib/strip-special-chars.js
@@ -1,0 +1,18 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+function stripSpecialChars(str) {
+  'use strict';
+  // normalize the strings to use double quotes, and
+  // replace newlines so that styles aren't found on
+  // multiple lines
+  if (typeof str === 'string' && str.length > 0) {
+    str = str.replace(/'/g, '"');
+    str = str.replace(/""/g, '');
+    str = str.replace(/\n/g, '');
+    return str.replace(/(\\)/g, '');
+  }
+}
+
+module.exports = stripSpecialChars;
--- a/grunttasks/po2json.js
+++ b/grunttasks/po2json.js
@@ -61,12 +61,12 @@ module.exports = function (grunt) {
       }
     },
     all: {
-      src: ['locale/**/**/*.po'],
-      dest: '.tmp/l10n/'
+      src: ['<%= config.strings_src %>/**/**/*.po'],
+      dest: '<%= config.tmp %>/l10n/'
     },
     template: {
-      src: ['locale/templates/**/*.pot'],
-      dest: '.tmp/l10n'
+      src: ['<%= config.strings_src %>/templates/**/*.pot'],
+      dest: '<%= config.tmp %>/l10n'
     }
   });
 };
--- a/grunttasks/po2json.js
+++ b/grunttasks/po2json.js
@@ -1,0 +1,74 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// grunt task to create .json files out of .po files.
+// .po files are expected to already be downloaded.
+
+var path = require('path');
+var i18n = require('i18n-abide');
+
+module.exports = function (grunt) {
+  grunt.config('po2json', {
+    options: {
+      format: 'raw',
+      fuzzy: true,
+      output_filename: function (file) { //eslint-disable-line camelcase
+
+        /**
+         * the files are stored in the locale subdirectory with a directory
+         * structure of:
+         * locale/
+         *       <locale_name>/
+         *                    LC_MESSAGES/
+         *                               server.po
+         *                               client.po
+         *
+         * Each locale is stored in its own subdirectory in the output i18n
+         * directory.
+         **/
+        var matches = /^locale\/([^\/]+)\/LC_MESSAGES\/(.*)$/.exec(file);
+
+        // In order to make sure the locale is in a form that i18n-abide expects
+        // we convert it from locale to langauge back to locale.
+        var locale = i18n.normalizeLocale(matches[1]);
+        var filename = matches[2];
+        if (filename === 'server.po') {
+          filename = 'messages.json';
+        } else {
+          // get rid of the .po extension, replace with .json
+          filename = path.basename(filename, '.po') + '.json';
+        }
+        return locale + '/' + filename;
+      },
+      output_transform: function (data) { //eslint-disable-line camelcase
+        // write the first translation only, ignore pluralization.
+        var isArray = function (item) {
+          return Object.prototype.toString.call(item) === '[object Array]';
+        };
+
+        var transformed = {};
+        for (var msgid in data) {
+          var translation = data[msgid];
+          if (isArray(translation) && translation.length >= 2) {
+            translation = translation[1];
+          }
+
+          transformed[msgid] = translation;
+        }
+
+        return transformed;
+      }
+    },
+    all: {
+      src: ['locale/**/**/*.po'],
+      dest: '.tmp/l10n/'
+    },
+    template: {
+      src: ['locale/templates/**/*.pot'],
+      dest: '.tmp/l10n'
+    }
+  });
+};
+
+
--- a/grunttasks/scan-email-translations.js
+++ b/grunttasks/scan-email-translations.js
@@ -3,113 +3,12 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 module.exports = function (grunt) {
+  'use strict';
   grunt.registerTask('scan-email-translations', 'Scan the generated emails for invalid style tags', [
     'generate-valid-styles',
     'po2json',
+    'l10n-json-to-html',
     'verify-styles'
     ]
   );
-
-  grunt.registerMultiTask('verify-styles', 'Verify whether the translated styles are valid', function (){
-    var options = this.options({
-      validStyles: []
-    });
-    var styleArray = options.validStyles;
-
-    var fileArray = this.files[0].src;
-
-    var hasErrors = 0;
-    var styleRegex = /style="[\s\S]*?"/ig;
-
-    fileArray.forEach(function (src) {
-      var json = grunt.file.readJSON(src);
-      var styleList = [];
-      for(var key in json) {
-        if(key.indexOf('style') > 0) {
-          styleList.push(key);
-        }
-      }
-      // Check if there were any styles found, .match() might return null
-      if (styleList !== null) {
-        styleList = styleList.map(stripSpecialChars);
-        // retrieve the style attribute and its contents and store
-        styleList = styleList.map(function (str) {
-          if (styleRegex.test(str)) {
-            str = str.match(styleRegex)[0];
-            return str;
-          }
-        });
-        styleList.forEach(function (style) {
-          if(style !== undefined){
-            if (styleArray.indexOf(style) === -1) {
-              grunt.log.error('Found a style which should not be there', style, src);
-              hasErrors = hasErrors + 1;
-            }
-          }
-        });
-      }
-    });
-    if (hasErrors !== 0) {
-      grunt.fail.warn('Found ' + hasErrors + ' ' + grunt.util.pluralize(hasErrors, 'error/errors') + ' in ' + fileArray.length + ' files');
-    } else {
-      grunt.log.writeln('Checked ' + fileArray.length + ' files for valid styles and found nothing wrong.');
-    }
-  });
-
-  grunt.config('verify-styles', {
-    dist: {
-      options: {
-        validStyles: grunt.file.readJSON('.tmp/validStyles.json')
-      },
-      files: [
-        {
-          expand: false,
-          src: [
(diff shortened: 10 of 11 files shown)


grunttasks/xss-mitigate.js
```diff
@@ -0,0 +1,85 @@
...
+          // On encountering any illegal tag, flag it
+          grunt.log.error('%s: INVALID TAGS found <%s>', src, tag);
+          hasErrors += 1;
+          errorFound = true;
+          return '';
+        },
+        onIgnoreTagAttr: function (tag, name, value, isWhiteAttr) {
+          // On encountering any illegal attribute, flag it
+          grunt.log.error('%s: INVALID TAG ATTRIBUTES FOUND <%s %s="">', src, tag, name);
+          hasErrors += 1;
```
pdehaan: **PROTIP:** Actually, I found [`this.errorCount`](http://gruntjs.com/api/inside-tasks#this.errorcount) in Grunt yesterday which may be useful here:

> The number of `grunt.log.error()` calls that occurred during this task. This can be used to fail a task if errors were logged during the task.

That way you don't need `hasErrors` or `errorFound` at all and you can just use `if (this.errorCount !== 0) {` below, along with `grunt.fail.warn('Found ' + this.errorCount + ' ...');`

grunttasks/xss-mitigate.js
```diff
@@ -0,0 +1,85 @@
...
+          strong: []
+        },
+        onTagAttr: function (tag, name, value, isWhiteAttr) {
+          // On encountering any tag, check if the tag has a valid attribute.
+          // If the attribute is valid, check for its value, flag if illegal
+          // Illegal includes any mention of the string 'javascript', case insensitive
+          // Templated strings are allowed i.e strings of the form '%(termsURI)s'
+          // If these are considered to be not secure, we should change
+          if (name !== 'style'){
+            if (isWhiteAttr && xss.safeAttrValue(tag, name, value) === '' && ! templateRegex.test(value)) {
```
pdehaan: Not sure if you want to combine this nested `if`.
Or if we need the extra logic for the `style` attribute, per the workaround in https://github.com/leizongmin/js-xss/issues/36#issuecomment-126114445

grunttasks/xss-mitigate.js
TDA: whoa, thanks Peter! will change!

rfk:
> feat(client): remove malicious code

It's a nit, but a casual reader would find this commit message quite alarming - it sounds like the commit literally removes some malicious code that we discovered lying around in our repo! :-)

Would you mind changing it to something a little milder, like "sanitize HTML in l10n translations"?

TDA:
@rfk Sure thing! Changed!

rfk:
Is this ready for a final r? and merge?

TDA:
@rfk Yes it is :)

rfk:
I recall @shane-tomlinson volunteering to sheppherd this through to a final merge, assigning to him for review or further delegation.

grunttasks/xss-mitigate.js
```diff
@@ -0,0 +1,78 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var xss = require('xss');
+
+module.exports = function (grunt) {
+  grunt.registerTask('xss-flag', 'Run the XSS parser on the generated l10n files', [
+    'l10n-json-to-html',
+    'xss-parse'
+    ]
```
shane-tomlinson: Nit, the `]` at the same indentation level as the above strings is strange.

grunttasks/json2html.js
```diff
@@ -0,0 +1,48 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var path = require('path');
+
+// Four steps are performed to lint the files:
+//  1. Clean the directories
+//  2. Generate the JSON files from the PO files
+//  3. The JSON files are then written out to temporary HTML files
+//  4. The HTML files are linted
```
shane-tomlinson: Can the HTML linter just take the JSON strings instead of this intermediate representation of HTML files?

grunttasks/json2html.js
```diff
@@ -0,0 +1,48 @@
...
+
+module.exports = function (grunt) {
+  grunt.registerMultiTask('l10n-json-to-html', 'Convert l10n JSON files to HTML', function () {
+    var files = 0;
+    this.files.forEach(function (file) {
+      var content = '';
+      var src = file.src[0];
+      var pathname = src.split('/');
+      var locale = grunt.file.readJSON(src);
```
shane-tomlinson: This could use a better name. It's not really a locale, but the locale's strings.

grunttasks/json2html.js
```diff
@@ -0,0 +1,48 @@
...
+      for (var val in locale) {
+        var value = locale[val];
+        if (typeof value !== 'object' && value !== '' && isNaN(value)) {
+          content += value.toString() + '\n\n';
+        }
+      }
+      if (content !== '') {
+        var tmpDir = '.tmp/html';
+        grunt.file.write(path.join(tmpDir, pathname[1], pathname[2], pathname[3].replace('json', 'html')), content);
```
shane-tomlinson: This makes a huge assumption about the directory structure. If the directory structure changes, this will break. Can [path.resovle](https://nodejs.org/api/path.html#path_path_resolve_from_to) be used here?

grunttasks/json2html.js
```diff
@@ -0,0 +1,48 @@
...
+    var files = 0;
+    this.files.forEach(function (file) {
+      var content = '';
+      var src = file.src[0];
+      var pathname = src.split('/');
+      var locale = grunt.file.readJSON(src);
+      for (var val in locale) {
+        var value = locale[val];
+        if (typeof value !== 'object' && value !== '' && isNaN(value)) {
+          content += value.toString() + '\n\n';
```
shane-tomlinson: Is content the HTML content? If so, the name should convey that.

grunttasks/json2html.js
```diff
@@ -0,0 +1,48 @@
...
+  grunt.registerMultiTask('l10n-json-to-html', 'Convert l10n JSON files to HTML', function () {
+    var files = 0;
+    this.files.forEach(function (file) {
+      var content = '';
+      var src = file.src[0];
+      var pathname = src.split('/');
+      var locale = grunt.file.readJSON(src);
+      for (var val in locale) {
+        var value = locale[val];
+        if (typeof value !== 'object' && value !== '' && isNaN(value)) {
```
shane-tomlinson: Do we ever expect *anything* other than strings? If so, we may have deeper problems. What happens if a value is not an object, or hits true on `isNaN`, does that mean it's unchecked? I think we should count that as an error to be safe.

grunttasks/json2html.js
```diff
@@ -0,0 +1,48 @@
...
+//  3. The JSON files are then written out to temporary HTML files
+//  4. The HTML files are linted
+
+module.exports = function (grunt) {
+  grunt.registerMultiTask('l10n-json-to-html', 'Convert l10n JSON files to HTML', function () {
+    var files = 0;
+    this.files.forEach(function (file) {
+      var content = '';
+      var src = file.src[0];
+      var pathname = src.split('/');
```
shane-tomlinson: This is not cross-platform compatible, Windows machines use `\`. Split on [path.sep](https://nodejs.org/api/path.html#path_path_sep).

grunttasks/json2html.js
```diff
@@ -0,0 +1,48 @@
...
+      var locale = grunt.file.readJSON(src);
+      for (var val in locale) {
+        var value = locale[val];
+        if (typeof value !== 'object' && value !== '' && isNaN(value)) {
+          content += value.toString() + '\n\n';
+        }
+      }
+      if (content !== '') {
+        var tmpDir = '.tmp/html';
```
shane-tomlinson: Paths should be created using [path.join](https://nodejs.org/api/path.html#path_path_join_path1_path2) so they are cross-platform compatible.

grunttasks/json2html.js
shane-tomlinson: Is there a `clean` task that removes old/stale files?

grunttasks/xss-mitigate.js
```diff
@@ -0,0 +1,78 @@
...
+  );
+
+  grunt.registerMultiTask('xss-parse', 'Parse HTML files to check for XSS vectors', function () {
+    var templateRegex = /\%\(\w+\)s/;
```
shane-tomlinson: What does this look for? `template` is a very overloaded word here.

grunttasks/xyz.js
```diff
@@ -0,0 +1,24 @@
+var xss = require('xss');
```
shane-tomlinson: What does this file do?

grunttasks/xss-mitigate.js
```diff
@@ -0,0 +1,78 @@
...
+    }
+  });
+
+  grunt.config('xss-parse', {
+    dist: {
+      files: [
+        {
+          expand: true,
+          src: [
+            '.tmp/html/l10n/*/*.html'
```
shane-tomlinson: `/*/`=>`/**/` to avoid assumptions about the directory structure.

grunttasks/xss-mitigate.js
```diff
@@ -0,0 +1,78 @@
...
+      // whitelist of allowed tags and attributes, nothing else is allowed
+      var errorFound = false;
+      var options = {
+        whiteList: {
+          a: ['href', 'id', 'style'],
+          em: [],
+          span: ['id', 'tabindex'],
+          strong: []
+        },
+        onTagAttr: function (tag, name, value, isWhiteAttr) {
```
shane-tomlinson: Can you rename `isWhiteAttr` to `isAllowedAttr`?

grunttasks/xss-mitigate.js
```diff
@@ -0,0 +1,78 @@
...
+        },
+        onIgnoreTagAttr: function (tag, name, value, isWhiteAttr) {
+          // On encountering any illegal attribute, flag it
+          grunt.log.error('%s: INVALID TAG ATTRIBUTES FOUND <%s %s="">', src, tag, name);
+          errorFound = true;
+          return '';
+        }
+      };
+      contents = xss(contents, options);
+      if (errorFound) {
```
shane-tomlinson: Can you rename `errorFound` to `isErrorFound`

grunttasks/xss-mitigate.js
```diff
@@ -0,0 +1,78 @@
...
+        // if an error was found in this file, write it back out.
+        grunt.file.write(src, contents);
+      }
+    });
+    if (this.errorCount !== 0) {
```
shane-tomlinson: Where does `this.errorCount` come from?

grunttasks/xss-mitigate.js
```diff
@@ -0,0 +1,78 @@
...
+          span: ['id', 'tabindex'],
+          strong: []
+        },
+        onTagAttr: function (tag, name, value, isWhiteAttr) {
+          // On encountering any tag, check if the tag has a valid attribute.
+          // If the attribute is valid, check for its value, flag if illegal
+          // Illegal includes any mention of the string 'javascript', case insensitive
+          // Templated strings are allowed i.e strings of the form '%(termsURI)s'
+          // If these are considered to be not secure, we should change
+            if (name !== 'style' && isWhiteAttr && xss.safeAttrValue(tag, name, value) === '' && ! templateRegex.test(value)) {
```
shane-tomlinson: IIUC, a string could be formed that is invalid, but just because it has a template string, it's allowed through?

shane-tomlinson: What I'm worried about is somebody doing something like:

`this string %(contains_a_template)s' onclick="alert(1)"`

shane-tomlinson: A bit of related light reading: http://blogs.technet.com/b/johnla/archive/2015/04/26/defenders-think-in-lists-attackers-think-in-graphs-as-long-as-this-is-true-attackers-win.aspx

shane-tomlinson: Ah, I see, only the attribute value is checked. Still, it leaves open room for something like:

`this string contains javascript <a href="javascript:alert(1);%(template)s"> but is not checked`

grunttasks/xss-mitigate.js
```diff
@@ -0,0 +1,78 @@
...
+              grunt.log.error('%s: INVALID VALUE FOR ATTRIBUTE <%s %s="%s">', src, tag, name, value);
+              return '';
+            }
+        },
+        onIgnoreTag: function (tag, html, options) {
+          // On encountering any illegal tag, flag it
+          grunt.log.error('%s: INVALID TAGS found <%s>', src, tag);
```
shane-tomlinson: Can you change `INVALID TAGS` to `UNEXPECTED TAG`

grunttasks/xss-mitigate.js
```diff
@@ -0,0 +1,78 @@
...
+          errorFound = true;
+          return '';
+        },
+        onIgnoreTagAttr: function (tag, name, value, isWhiteAttr) {
+          // On encountering any illegal attribute, flag it
+          grunt.log.error('%s: INVALID TAG ATTRIBUTES FOUND <%s %s="">', src, tag, name);
```
shane-tomlinson: Can you change `INVALID TAG ATTRIBUTES` to `UNEXPECTED ATTRIBUTE`

grunttasks/xss-mitigate.js
```diff
@@ -0,0 +1,78 @@
...
+        },
+        onTagAttr: function (tag, name, value, isWhiteAttr) {
+          // On encountering any tag, check if the tag has a valid attribute.
+          // If the attribute is valid, check for its value, flag if illegal
+          // Illegal includes any mention of the string 'javascript', case insensitive
+          // Templated strings are allowed i.e strings of the form '%(termsURI)s'
+          // If these are considered to be not secure, we should change
+            if (name !== 'style' && isWhiteAttr && xss.safeAttrValue(tag, name, value) === '' && ! templateRegex.test(value)) {
+              grunt.log.error('%s: INVALID VALUE FOR ATTRIBUTE <%s %s="%s">', src, tag, name, value);
+              return '';
```
shane-tomlinson: Why return `''`? What is the difference between that and returning `null`, `undefined`, `false`, or `0`?

TDA: Returning '' ensures that the string is replace with ''. The return value directly replaces the original string in this module.

TDA:
@shane-tomlinson @vladikoff  this PR.

grunttasks/json2html.js
TDA: the xss module does not run on json unfortunately. it needs to form a dom tree with the html tags internally so.

grunttasks/json2html.js
TDA: fixed.

grunttasks/json2html.js
TDA: will add a clean task.

grunttasks/json2html.js
TDA: added.

grunttasks/json2html.js
TDA: the json has some metadata objects, so those dont need to be converted to html, these will not be generated by our translators either so.
I dont understand the concern about the second part, if its not an object and hits true on isNaN, it *will* be checked.

grunttasks/json2html.js
TDA: added.

grunttasks/json2html.js
TDA: added.

grunttasks/xss-mitigate.js
TDA: it looks for usage of templates like so:
%(sampleURI)
What do you think is a better name?

grunttasks/xss-mitigate.js
TDA: done.

grunttasks/xss-mitigate.js
pdehaan: @shane-tomlinson It's a Grunt task thing: http://gruntjs.com/api/inside-tasks#this.errorcount

> _"The number of `grunt.log.error()` calls that occurred during this task. This can be used to fail a task if errors were logged during the task."_

grunttasks/xss-mitigate.js
TDA: I thought I would stick to what the module uses, so that it would be easier for someone who checks the code later on.
https://github.com/leizongmin/js-xss/blob/master/README.md#customize-the-handler-function-for-attributes-of-matched-tags

grunttasks/xss-mitigate.js
TDA: done.

grunttasks/xss-mitigate.js
TDA: its a grunt shortcut that @pdehaan found :D really handy!

grunttasks/xss-mitigate.js
TDA: changed the regex to make sure the template string is the ONLY value in the attr-value. 
Good catch, this is one of those attack planes that cannot be checked without another whitelist of templateURIs
@shane-tomlinson

grunttasks/xss-mitigate.js
TDA: done.

grunttasks/xss-mitigate.js
TDA: Changed, but I would reckon Invalid *does* mean malformed, as this refers to a attr like 'src', while line 36 checks for the validity of the value of that 'src' attr.

grunttasks/generate-valid-styles.js
```diff
@@ -0,0 +1,42 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var path = require('path');
+var stripSpecialChars = require('./strip-special-chars');
+var VALID_STYLES_FILE_NAME = 'validStyles.json';
```
shane-tomlinson: Include a common config task with the filename declared within. For an example, see how [the content server does it](https://github.com/mozilla/fxa-content-server/blob/master/grunttasks/yeoman.js). Rename the file from camelCase to either snake_case or use-a-dash.

grunttasks/generate-valid-styles.js
```diff
@@ -0,0 +1,42 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var path = require('path');
+var stripSpecialChars = require('./strip-special-chars');
```
shane-tomlinson: Having a non-task js file in with the rest of the tasks is a bit confusing, maybe put this under a `lib` subdirectory.

grunttasks/generate-valid-styles.js
shane-tomlinson: Another reason to use a configuration task is to consolidate declarations of this name, since [it's currently duplicated here](https://github.com/mozilla/fxa-content-server-l10n/pull/63/files#diff-48bf94fdac401ff9cacb51d8adc46841R7).

grunttasks/generate-valid-styles.js
```diff
@@ -0,0 +1,42 @@
...
+    // iterate through .pot file and extract the styles
+    this.files[0].src.forEach(function (src) {
+      var contents = grunt.file.read(src);
+      if (validStylesRegex.test(contents)) {
+        validStylesArray = validStylesArray.concat(contents.match(validStylesRegex));
+      }
+    });
+    validStylesArray = validStylesArray.map(stripSpecialChars);
+    // write out the styles to a temporary JSON file
+    grunt.file.write(path.join(this.files[0].dest, VALID_STYLES_FILE_NAME), JSON.stringify(validStylesArray));
```
shane-tomlinson: What is the full path of the json file? Does this add the temporary json file in with the rest of the .po/.pot files? Temporary files should not pollute the original directory tree, let's use a subdirectory that can be cleaned for this purpose. In the content server, we use `.tmp`

shane-tomlinson: Ah, I see. That's what you are doing.

grunttasks/json2html.js
shane-tomlinson: > the xss module does not run on json unfortunately. it needs to form a dom tree with the html tags internally so.

Can you add that to the comment so the knowledge is preserved?

grunttasks/json2html.js
```diff
@@ -0,0 +1,50 @@
...
+      var pathname = src.split(path.sep);
+      var localeStrings = grunt.file.readJSON(src);
+      for (var val in localeStrings) {
+        var value = localeStrings[val];
+        if (typeof value !== 'object' && value !== '' && isNaN(value)) {
+          HTMLcontent += value.toString() + '\n\n';
+        }
+      }
+      if (HTMLcontent !== '') {
+        var tmpDir = '.tmp';
```
shane-tomlinson: I see `.tmp` sprinked around, that should be moved into the configuration task I mentioned earlier.

grunttasks/verify-styles.js
```diff
@@ -0,0 +1,70 @@
...
+              str = str + matches[style];
+              return str;
+            }
+          }
+        });
+      }
+      generatedStylesList.forEach(function (style) {
+          if(style && style.length > 0){
+            if (validStylesArray.indexOf(style) === -1) {
+              grunt.log.error('Found a style which should not be there', style, src);
```
shane-tomlinson: Nit: Some spacing could be added. The output currently looks like:

```
 % grunt scan-email-translations
Running "generate-valid-styles:dist" (generate-valid-styles) task
Generated styles from 2 files

Running "po2json:all" (po2json) task

Running "po2json:template" (po2json) task

Running "verify-styles:dist" (verify-styles) task
>> Found a style which should not be there For more information, please visit <a href="https://support.mozilla.org" style="color: #0095dd; text-decoration: none;">Mozilla Support</a>style="color: #0095dd; text-decoration: none;" .tmp/l10n/be/messages.json
```

grunttasks/verify-styles.js
```diff
@@ -0,0 +1,70 @@
...
+    }
+  });
+
+    grunt.config('verify-styles', {
+    dist: {
+      files: [
+        {
+          expand: false,
+          src: [
+            '.tmp/l10n/be/*.json'
```
shane-tomlinson: Do you want to only do `be`?

grunttasks/xss-mitigate.js
```diff
@@ -0,0 +1,77 @@
...
+var xss = require('xss');
+
+module.exports = function (grunt) {
+  grunt.registerTask('xss-flag', 'Run the XSS parser on the generated l10n files', [
+    'l10n-json-to-html',
+    'xss-parse'
+  ]);
+
+  grunt.registerMultiTask('xss-parse', 'Parse HTML files to check for XSS vectors', function () {
+    var templateRegex = /^\%\(\w+\)s$/;
```
shane-tomlinson: Can you rename `templateRegex` to `interpolatedValueRegex`?

grunttasks/xss-mitigate.js
shane-tomlinson: Can you break this line up? It scrolls off the right hand side of the GitHub review window. I often put each distinct test on its own line.

grunttasks/xss-mitigate.js
shane-tomlinson: ok.

grunttasks/xss-mitigate.js
```diff
@@ -0,0 +1,77 @@
...
+        onIgnoreTagAttr: function (tag, name, value, isWhiteAttr) {
+          // On encountering any illegal attribute, flag it
+          grunt.log.error('%s: UNEXPECTED TAG ATTRIBUTES FOUND <%s %s="">', src, tag, name);
+          isErrorFound = true;
+          return '';
+        }
+      };
+      contents = xss(contents, options);
+      if (isErrorFound) {
+        // if an error was found in this file, write it back out.
```
shane-tomlinson: I'm confused here, the source file is overwritten with the "clean" version? `contents` is already loaded from a generated HTML file, does writing out clean HTML serve any function later on?

shane-tomlinson:
@TDA - this is starting to look really good! I have added a few more questions and comments. I immediately receive an error when running `grunt scan-email-translations`, us that expected?

```bash
% grunt scan-email-translations
Running "generate-valid-styles:dist" (generate-valid-styles) task
Generated styles from 2 files

Running "po2json:all" (po2json) task

Running "po2json:template" (po2json) task

Running "verify-styles:dist" (verify-styles) task
>> Found a style which should not be there For more information, please visit <a href="https://support.mozilla.org" style="color: #0095dd; text-decoration: none;">Mozilla Support</a>style="color: #0095dd; text-decoration: none;" .tmp/l10n/be/messages.json
Warning: Found 1 error in 2 files Use --force to continue.

Aborted due to warnings.


Execution Time (2015-08-10 10:41:53 UTC)
loading tasks  276ms  ▇▇▇▇▇▇▇▇▇▇▇ 21%
po2json:all       1s  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 78%
Total 1.3s

 %
```

grunttasks/json2html.js
TDA: added.

grunttasks/xss-mitigate.js
TDA: changed :)

grunttasks/verify-styles.js
TDA: oops, was testing, forgot to push after that sorry!.

grunttasks/xss-mitigate.js
TDA: changed.

grunttasks/xss-mitigate.js
TDA: I believe @vladikoff asked me to add this for the content-server PR, should I remove it?

TDA:
@shane-tomlinson That was me testing, I manually added an error to the files so that they will throw an error, I will push an updated version soon.

grunttasks/xss-mitigate.js
vladikoff: Sorry, what's this for?

TDA: For sanitizing the generated html files. Do we need to write out the clean html back, or just log errors?

vladikoff: > For sanitizing the generated html files. Do we need to write out the clean html back, or just log errors?

If some bad html is found then the task should throw errors and notify us so we can remove it from the source.

TDA: All right, removing that part then.

TDA:
@shane-tomlinson Should be good now.

shane-tomlinson:
@TDA - I'm testing pretty rigorously now. With this patch, I receive no errors from either `grunt scan-email-translations` or `grunt xss-flag`:

```diff
diff --git a/locale/es/LC_MESSAGES/client.po b/locale/es/LC_MESSAGES/client.po
index b68bff7..5afcd30 100644
--- a/locale/es/LC_MESSAGES/client.po
+++ b/locale/es/LC_MESSAGES/client.po
@@ -111,7 +111,7 @@ msgstr "Se ha eliminado correctamente la cuenta"
 
 #: app/scripts/views/reset_password.js:82 app/scripts/views/sign_in.js:180
 msgid "Unknown account. <a href=\"/signup\">Sign up</a>"
-msgstr "Cuenta desconocida. <a href=\"/signup\">Regístrate</a>"
+msgstr "Cuenta desconocida. <a href=\"javascript:alert(1)\" onclick=\"javascript:alert('hey')" style=\"is this caught\">Regístrate</a>"
 
 #: app/scripts/views/settings.js:75
 msgid "Signed out successfully"
diff --git a/locale/es/LC_MESSAGES/server.po b/locale/es/LC_MESSAGES/server.po
index f01c768..0cc453a 100644
--- a/locale/es/LC_MESSAGES/server.po
+++ b/locale/es/LC_MESSAGES/server.po
@@ -115,7 +115,7 @@ msgid ""
 msgstr ""
 "Este email se ha generado automáticamente; si no has restablecido la "
 "contraseña de tu cuenta Firefox, deberías hacerlo inmediatamente en <a "
-"href='%(resetLink)s'>%(resetLink)s</a>."
+"href='%(resetLink)s' style='is this caught'>%(resetLink)s</a>."
 
 #: templates/password_reset.txt:3
 msgid ""
``` 

```bash
➜  fxa-content-server-l10n git:(sai) ✗ grunt scan-email-translations
Running "generate-valid-styles:dist" (generate-valid-styles) task
Generated styles from 2 files

Running "po2json:all" (po2json) task

Running "po2json:template" (po2json) task

Running "verify-styles:dist" (verify-styles) task
Checked 214 files for valid styles and found nothing wrong.

Done, without errors.


Execution Time (2015-08-11 14:24:10 UTC)
loading tasks       201ms  ▇▇▇▇▇▇▇▇▇▇ 20%
po2json:all         719ms  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 72%
verify-styles:dist   65ms  ▇▇▇ 7%
Total 998ms

➜  fxa-content-server-l10n git:(sai) ✗ grunt xss-flag
Running "po2json:all" (po2json) task

Running "po2json:template" (po2json) task

Running "l10n-json-to-html:dist" (l10n-json-to-html) task
>> 105 files converted from JSON to HTML

Running "xss-parse:dist" (xss-parse) task
Checked 105 files for XSS vulnerabilities

Done, without errors.


Execution Time (2015-08-11 14:24:18 UTC)
loading tasks           203ms  ▇▇▇▇▇▇▇▇ 19%
po2json:all             732ms  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 68%
l10n-json-to-html:dist  100ms  ▇▇▇▇ 9%
xss-parse:dist           37ms  ▇▇ 3%
Total 1.1s
```

You can see this is obviously invalid:
```
msgstr "Cuenta desconocida. <a href=\"javascript:alert(1)\" onclick=\"javascript:alert('hey')" style=\"is this caught\">Regístrate</a>"
```
That line has multiple problems. First, the href is a JS. Next, the onlcikc handler is not properly escaped, but the first part is javascript and will be run. The style should not be there.

In the server template:
```
"href='%(resetLink)s' style='is this caught'>%(resetLink)s</a>."
```

That style shouldn't be there, but is let through.

grunttasks/verify-styles.js
```diff
@@ -0,0 +1,70 @@
...
+
+module.exports = function (grunt) {
+  'use strict';
+  grunt.registerMultiTask('verify-styles', 'Verify whether the translated styles are valid', function (){
+    var validStylesFile = path.join(this.files[0].dest, grunt.config.process('<%= yeoman.validStylesFile %>'));
+    var validStylesArray = grunt.file.readJSON(validStylesFile);
+    var filesToBeChecked = this.files[0].src;
+
+    var hasErrors = 0;
+    var generatedStylesRegex = /style=('|")[\S\s]*?('|")/ig;
```
shane-tomlinson: I forgot, in HTML5, the quotes are strictly optional. A style can be considered valid w/o the quotes.

```
style=backbroundImage:none
```

Instead of using regular expressions to check styles, we may want to use the full html parser in https://github.com/mozilla/fxa-content-server-l10n/pull/63/files#diff-89412030af742d9041e0c356ccf997c7R5, moving the checks from this file there.

grunttasks/verify-styles.js
```diff
@@ -0,0 +1,70 @@
...
+
+    grunt.config('verify-styles', {
+    dist: {
+      files: [
+        {
+          expand: false,
+          src: [
+            '<%= yeoman.tmp %>/l10n/**/*.json'
+          ],
+          dest: '<%= yeoman.tmp %>'
```
shane-tomlinson: We should definitely not use `dest` to indicate where the validStylesFile is. The convention for `dest` is the destination filename/directory of the output of the current task.

vladikoff:
@TDA Didn't we also have some test code for these tasks? I thought you asked me earlier about testing these

grunttasks/generate-valid-styles.js
```diff
@@ -0,0 +1,41 @@
...
+  'use strict';
+   grunt.registerMultiTask('generate-valid-styles', 'Generate the list of valid styles for translation', function () {
+    var validStylesRegex = /style=('|")[\s\S]*?('|")/ig;
+    var validStylesArray = [];
+
+    // iterate through .pot file and extract the styles
+    this.files[0].src.forEach(function (src) {
+      var contents = grunt.file.read(src);
+      if (validStylesRegex.test(contents)) {
+        validStylesArray = validStylesArray.concat(contents.match(validStylesRegex));
```
shane-tomlinson: I was digging into this pretty deeply today, it only seems to extract the first style in the file. I think we are going to have to use RegExp.exec and a while loop, [as shown in this MDN example](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#Finding_successive_matches).

grunttasks/generate-valid-styles.js
```diff
@@ -0,0 +1,41 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var path = require('path');
+var stripSpecialChars = require('./lib/strip-special-chars');
+
+module.exports = function (grunt) {
+  'use strict';
+   grunt.registerMultiTask('generate-valid-styles', 'Generate the list of valid styles for translation', function () {
+    var validStylesRegex = /style=('|")[\s\S]*?('|")/ig;
```
shane-tomlinson: We should probably capture the inner style contents instead of the entire attribute.

grunttasks/xss-mitigate.js
```diff
@@ -0,0 +1,76 @@
...
+          grunt.log.error('%s: UNEXPECTED TAGS found <%s>', src, tag);
+          return '';
+        },
+        onIgnoreTagAttr: function (tag, name, value, isWhiteAttr) {
+          // On encountering any illegal attribute, flag it
+          grunt.log.error('%s: UNEXPECTED TAG ATTRIBUTES FOUND <%s %s="">', src, tag, name);
+          return '';
+        }
+      };
+      contents = xss(contents, options);
```
shane-tomlinson: This seems to not care about broken tags. For example, I have generated this HTML snippet which goes undetected:

```html
Cuenta desconocida. <a href="javascript(1)" onclick="javascript:alert('hey')

Desconectado correctamente

La cuenta ya existe. <a href="/signin">Identifícate</a>
```

grunttasks/yeoman.js
```diff
@@ -0,0 +1,17 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = function (grunt) {
```
shane-tomlinson: Let's rename both this file and the config option `config`, `yeoman` is a product that we used as the initial scaffolding for the content server, which generated the file for us. The content server name is purely because we haven't updated it, let's do the right thing here.

grunttasks/generate-valid-styles.js
TDA: @shane-tomlinson https://www.dropbox.com/s/7qgdk01hwnoifxh/Screenshot%202015-08-11%2015.50.03.png?dl=0
I am able to see that it finds all the styles in the .pot files. Or am i misunderstanding the problem?

grunttasks/generate-valid-styles.js
TDA: So this is following up from another thread where we decided to extract the entire attribute instead of just the contents.
I think @vladikoff wanted us to do that.

grunttasks/xss-mitigate.js
TDA: @shane-tomlinson You are perfectly right, the module doesnt seem to care about broken tags!!! whats worse, it ignores anything after a broken unclosed tag :( :( I am filing an issue with the xss module people for this.

TDA:
@shane-tomlinson The part about styles getting missed even when mentioned in the .po files is now fixed. Turns out that the json parsing i was doing locally in verify-styles only looked at the keys(which are actually the untranslated stuff, duh me :( ).
So instead of reproducing the same stuff as json2html, i just added that to the scan-email-translations, and modified verify-styles to read html instead of json. Lesser code, and will catch the styles for sure.

I am not sure on what to do about malformed markup that could potentially escape through the xss module. Surely, a combination of this with 'html-lint' from content server would prevent those? https://github.com/mozilla/fxa-content-server/pull/2726

Seeing as if anything passes that, then they should be parsed by the xss module, so one of them has to throw an error? ( they are basically like an XOR here). I feel those should be enough to pass almost all attack vectors, but then, one can never be too sure.

vladikoff:
Let's figure out what is blocking this and fix this up. Making a checkbox list with a TODO might help

TDA:
@vladikoff The issue that was present was fixed by the owner of that repo, so I am gonna update the package.json file. I dont think any other issues are present in this.

shane-tomlinson:
Let's hold off on merging this. There were still several edge cases and simplifications that I was slowly working through in the weeks before I went on vacation.

shane-tomlinson:
Closing and opening a new PR

PR #68, opened 2015-08-27
merged server strings for train 45

dannycoates:
Contains the new email strings for https://github.com/mozilla/fxa-auth-mailer/pull/72


vladikoff:
Conflicts! :scream:

dannycoates:
moving target :dart:

mathjazz:
Hey, just checking why these strings got merged without a notice? Updating Verbatim...

@dannycoates @zaach

mathjazz:
I'd suggest against putting this into localizable strings, because it's confusing localizers, some of them will translate these words:
<a href='%(supportUrl)s' style='color: #0095dd; text-decoration: none;'>

Also, do we really need all these string:
- Activate Now
- Activate now
- Activate now:

rfk:
> some of them will translate these words

@mathjazz do you mean the "style" declarations specifically?

mathjazz:
@rfk I mean we should simply use something like this instead:
```For more information, please visit <a %(supportUrl)s>Mozilla Support</a>```

Less markup means less possibilities for localizers to break stuff. ;)

Also, we already have this string:
```For more information, please visit %(supportUrl)s```

zaach:
Perhaps we can do something clever like `For more information, please visit <a %(linkAttributes)s>Mozilla Support</a>` and inject all of the attributes rather than just the URL.

The problem with reusing `For more information, please visit %(supportUrl)s` is that it doesn't include `Mozilla Support`, which I suppose may affect how it is translated.

PR #69, opened 2015-09-09
feat(build): scan translations for invalid attributes

### LINKED ISSUE #61: Move Grunt XSS tasks to the l10n repo

shane-tomlinson:
* The allowed list of attributes is generated from the .pot files in locale/templates/*
* A list of expected attributes and values are saved.
* When xss-parse is run, all attributes in the translated files are checked
  against the allowed list.
* no intermediate representations are used, extracting allowed attributes and examining translations are all done against the original .po and .pot files.

fixes #61


shane-tomlinson:
@zaach, @vladikoff, and @rfk - r?

shane-tomlinson:
fixes #61

vladikoff:
@shane-tomlinson does it make sense to add Travis CI with this? so it scans when there is a new build?

shane-tomlinson:
> @shane-tomlinson does it make sense to add Travis CI with this? so it scans when there is a new build?

It does.

.travis.yml
```diff
@@ -0,0 +1,29 @@
+language: node_js
+
+node_js:
+  - "0.12"
```
vladikoff: awww yeaaaaaa

shane-tomlinson:
Not ready yet! I just found some issues.

.travis.yml
pdehaan: what, no Node v4 support?

vladikoff: We can use
```
node_js:
  -  "4"
```

rfk:
@shane-tomlinson did we decide on whether it was safe for the logic to live in this repo, where push access may be less tightly controlled than other project repos?

rfk:
I really like how simple this seems.  Do you have thoughts on hooking it up to some automated testing so we can be sure it continues to work after we stop paying attention to it?

shane-tomlinson:
> I really like how simple this seems. Do you have thoughts on hooking it up to some automated testing so we can be sure it continues to work after we stop paying attention to it?

As part of this PR, it's hooked up to travis. :D

shane-tomlinson:
> As part of this PR, it's hooked up to travis. 

We should probably have travis failures notify us directly.

shane-tomlinson:
> @shane-tomlinson did we decide on whether it was safe for the logic to live in this repo, where push access may be less tightly controlled than other project repos?

Yeah, I want to remove the from this repo and into its own repo, and sooner rather than later.

rfk:
> thoughts on hooking it up to some automated testing

I also mean, how do we test that this code itself is working, and will actually detect problems if they show up in the translations?

shane-tomlinson:
> I also mean, how do we test that this code itself is working, and will actually detect problems if they show up in the translations?

I'm working on a branch for just that problem.

shane-tomlinson:
Ref https://github.com/mozilla/grunt-l10n-lint, where the lint logic now lives.

rfk:
Is this ready to have "WIP" removed and find a formal reviewer?

shane-tomlinson:
> Is this ready to have "WIP" removed and find a formal reviewer?

Yes.

.travis.yml
```diff
@@ -0,0 +1,30 @@
...
+cache:
+  directories:
+    - node_modules
+
+notifications:
+  irc:
+    channels:
+      - 'irc.mozilla.org#fxa-bots'
+    use_notice: false
+    skip_join: false
```
rfk: This seems important enough what we should add email notifications on failure, they'd be easily missed in IRC.

package.json
```diff
@@ -13,5 +13,11 @@
   "license": "MPL-2.0",
   "bugs": {
     "url": "https://github.com/mozilla/fxa-content-server-l10n/issues"
+  },
+  "devDependencies": {
+    "grunt": "0.4.5",
+    "grunt-l10n-lint": "0.0.0",
```
rfk: I'm surprised and impressed that npm lets you register this as a package version....

rfk:
r+ on this PR; I've also added the grunt-l10n-lint repo to our waffleboard for any follow-up bugs/comments.

.travis.yml
shane-tomlinson: > This seems important enough what we should add email notifications on failure, they'd be easily missed in IRC.

Added.

PR #70, opened 2015-09-16
Remove unexpected `em` tags in the Italian translation.

shane-tomlinson:
`Scarica Firefox per dispositivi <em>mobile</em>.` =>
`Scarica Firefox per dispositivi mobile.`

@mathjazz - could you review this? I found the extra tag while working on a linting tool (see https://github.com/mozilla/grunt-l10n-lint)


mathjazz:
r+, nice catch!

PR #77, opened 2015-10-15
merge strings for train 48

shane-tomlinson:
@mathjazz - only one new string this train! I wasn't sure if I needed to do a PR this time, or whether we should start pushing directly to master.


mathjazz:
r+!

PR #78, opened 2015-10-15
Updated Polish (pl) localization of Firefox Accounts for train 48

splewako:


shane-tomlinson:
@mathjazz - are we good here?

mathjazz:
LGTM

splewako:
Let's change that in the next update, we should also check other projects for `autoryz*`

PR #79, opened 2015-10-17
[pt-PT] Localization updates;

cesperanc:


shane-tomlinson:
@mathjazz - we good to merge?

mathjazz:
Actually pt-PT should be just a symlink of pt, because they are the same thing and Verbatim only commits to pt, right @cesperanc?

cesperanc:
Yes, they are the same thing, with the exception of the "Language" tag (on the heading of the files) which seems to be different between those two directories. However the content is, exactly, the same.

mathjazz:
Good point @cesperanc!

@shane-tomlinson Does that line starting with "Language:" affect anything? I mean, what language will be served to users, who have browser language set to pt or pt-PT?

rfk:
@jrgm has previously requested that we change the `"Language:"` line to properly match the name of the file, although I don't recall the specific reason why.  It seems like good hygiene in any case.

If it's a straight copy, we should build a way to script up the logic in this repo rather than depending on PRs to keep it up to date.

shane-tomlinson:
> @shane-tomlinson Does that line starting with "Language:" affect anything? I mean, what language will be served to users, who have browser language set to pt or pt-PT?

AFAIK, the "Lanuage" line is not used anywhere. If users have their locale set to `pt-PT`, we'll ship them the normal `pt` locales. only `pt` and `pt-BR` are [enabled in production](https://github.com/mozilla/fxa-content-server/blob/master/server/config/production-locales.json#L35-L36). We can add `pt-PT` to the next train if @jrgm gives the green light.

@jrgm, thoughts?

mathjazz:
A slight correction. Most projects only keep pt-PT, which makes sense given that there's also pt-BR. And then probably use a symlink or some other mechanism to also serve pt-PT content to pt users.

I agree with @rfk - we should only keep pt-PT and avoid PRs for every change.

cesperanc:
Hi.

I am closing this PR because I've migrated the strings (with some minor updates) to pontoon and they should be pushed to this repository in a few minutes. However the question to serve pt-PT content to pt users still remains open.

Thank you

PR #83, opened 2015-10-27
Add Kazakh (kk) translation.

crayxt:
Added initial Kazakh translation.

### DIFF
--- a/locale/kk/LC_MESSAGES/client.po
+++ b/locale/kk/LC_MESSAGES/client.po
@@ -1,439 +1,445 @@
 msgid ""
 msgstr ""
-"Project-Id-Version: PACKAGE VERSION\n"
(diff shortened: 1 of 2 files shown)


seanmonstar:
@mathjazz whats the proper procedure with these?

rfk:
> whats the proper procedure with these?

In particular, do you want these to come in via pontoon rather than as standalone PRs?

mathjazz:
We suggest localizers to use our tools (e.g. Pontoon), because permission system, quality checks, translation memory and such.

That being said, Baurzhan is our kk localizer, so this is r+. You can usually check if PR is safe to take by looking at the L10n team wiki (https://wiki.mozilla.org/L10n:Teams:kk).

shane-tomlinson:
Thanks for the quick response @mathjazz, merged.

PR #86, opened 2015-10-30
New German strings for Firefox Accounts

koehlermichael:


koehlermichael:
Should have worked now via Pontoon.

PR #90, opened 2015-11-10
fix(l10n): convert sr to sr_Latn automatically

### LINKED ISSUE #82: Automate translation of "sr" to "sr_Latn"
As described in https://github.com/mozilla/fxa-content-server-l10n/issues/44#issuecomment-146425994

vladikoff:
Fixes #82


vladikoff:
This works well, but we need to figure out how to install `gettext` easily (that gives us `msgfilter`).

vladikoff:
@shane-tomlinson check it out ^

vladikoff:
@flodolo what's the easiest way to get `msgfilter` on developer machines (os x , ubuntu)?

flodolo:
On Linux you shouldn't have particular issues, `apt-get install gettext` should be enough.

On OS X it's a different story: first you need brew, then you need to install the `gettext`package but it won't give you an executable, so you need to call it as `/usr/local/opt/gettext/bin/msgfilter`.
Unfortunately I'm not that much of an expert with brew, I always do this kind of work in virtual machines running Linux.

vladikoff:
@shane-tomlinson Ready!

PR #96, opened 2015-11-13
fix(locale): copies locales that are not translated externally

### LINKED ISSUE #91: Add a right_before_production_deploy_script to copy locales
This script will do all the copying and should run right before prod deploy to get the most translations.
This is a bit of opposite to the "extract_strings", that one runs right after prod is built.

- [ ] Automatically copy/symlink "pt-PT" to "pt" https://github.com/mozilla/fxa-content-server-l10n/issues/81
- [ ] We might need to copy `hi-IN` to `hi`
- [ ] `sv_SE` => `sv`
- [ ] Remove all existing copy task in content-server  https://github.com/mozilla/fxa-content-server/blob/master/grunttasks/copy.js#L86 and copy them as part of this script

vladikoff:
Fixes #91

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,8 +13,7 @@ On a (different) regular basis, a SOURCE committer will update in the other dire
 
 ## Dependencies
 
-Install [gettext](https://www.gnu.org/software/gettext/).
-
+* Install [gettext](https://www.gnu.org/software/gettext/).
 OS X: 
 ```
 brew install gettext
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -23,17 +22,27 @@ brew link gettext --force # brew does not link the tools into /usr/local. Force
 
 Ubuntu: `sudo apt-get install gettext`
 
+The process to work with strings:
+
+* Make sure you installed dependencies and `msgfilter` (from `gettext`) command is available in your shell.
+* Have the fxa-auth-mailer and fxa-content-server repos at the directory level as fxa-content-server-l10n.
```
rfk: nit: "at the *same* directory level"?

scripts/before_prod_deploy.sh
```diff
@@ -0,0 +1,29 @@
...
+do
+param="$1"
+
+case $param in
+    --l10n-repo)
+    L10N_DIR="$2"
+    shift 2
+    ;;
+    *)
+    usage
```
rfk: `usage` is undefined?

rfk:
The code change here looks pretty small and solid (modulo nits) so I'm happy to r+ this from a mechanical standpoint, if you and @shane-tomlinson and previously agreed on the details of the approach.

scripts/before_prod_deploy.sh
```diff
@@ -0,0 +1,29 @@
...
+    ;;
+esac
+done
+
+# some locales must be copied
+# see details in https://github.com/mozilla/fxa-content-server-l10n/pull/79#issuecomment-149281761
+cp -r $L10N_DIR/locale/sv_SE/* $L10N_DIR/locale/sv
+cp -r $L10N_DIR/locale/hi_IN/* $L10N_DIR/locale/hi
+cp -r $L10N_DIR/locale/pt_PT/* $L10N_DIR/locale/pt
```
vladikoff: Add Serbian latin to this

vladikoff:
Thanks Ryan!

PR #101, opened 2016-01-11
fix(locales): adjust Language name for sr_Latn

jrgm:
This flipped back to 'sr' again. Setting it back again.

r? - @vladikoff


vladikoff:
@jrgm we need to make sure we run normalization before "release build" https://github.com/mozilla/fxa-content-server-l10n#locale-normalization

PR #126, opened 2016-05-27
fix(locale): adjust plural forms for lt

vladikoff:
cc @jrgm @jbuck


vladikoff:
cc @mathjazz @zygimantus this `or` statement was introduced only in server translations. This broke `lt` account creation due to parser errors. Could you explain how it got here? Is this some new format?

mathjazz:
That was a bug in Pontoon we fixed recently. Apparently we  didn't commit all the files for Firefox Accounts. Thanks for taking care of it!

PR #128, opened 2016-05-31
fix(scripts): Set the output width to 200 charcters

shane-tomlinson:
fixes mozilla/fxa-content-server#3795

@vladikoff - r?


vladikoff:
Tested it here, works !

PR #131, opened 2016-06-01
fix(l10n): remove old server.pot and use a fresh one each extraction

vladikoff:
@shane-tomlinson r?

Connects to https://github.com/mozilla/fxa-content-server/issues/3802

### DIFF
--- a/scripts/extract_strings.sh
+++ b/scripts/extract_strings.sh
@@ -66,7 +66,7 @@ check_folder $L10N_DIR
 
 set -x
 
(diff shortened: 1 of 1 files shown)


vladikoff:
It is already in gitignore here: https://github.com/mozilla/fxa-auth-mailer/blob/master/.gitignore#L2

if you extract strings and the old one is there it will lead to issues (such as old strings showing up)

vladikoff:
Fixes https://github.com/mozilla/fxa-content-server/issues/3802

PR #139, opened 2016-07-12
fix(ja): Fix the unexpected link.

shane-tomlinson:
Remove the /ja/ in the middle of the link, as well as the trailing slash.


vladikoff:
Awesome, thanks Shane!

PR #159, opened 2016-11-06
fix(git): Only ignore client.pot and server.pot in templates/LC_MESSAGES

### LINKED ISSUE #158: grunt l10n-lint fails
> grunt l10n-lint --verbose
....
> Reading locale/templates/LC_MESSAGES/client.pot...OK
> Reading locale/templates/LC_MESSAGES/messages.pot...OK
> Warning: Cannot convert undefined or null to object Use --force to continue.

This occurs using node 4. Also fails on travis: https://travis-ci.org/mozilla/fxa-content-server-l10n/jobs/172871222

shane-tomlinson:
fixes #158

@vladikoff - r?


TDA:
@shane-tomlinson Seems to be failing node4 with same error as before. Cached version maybe?

PR #194, opened 2017-04-13
merge strings for train 44511

bug-mirror:


vladikoff:
This is working but we are going to do string extraction on Friday, wooo!

PR #212, opened 2017-07-10
fix(es): Fix the broken link in es.

shane-tomlinson:
A space was missing between `a` and `%(escapedBackLinkAttrs)s`


vladikoff:
¡Si!

PR #229, opened 2017-09-29
merge strings for train 20339

bug-mirror:


vladikoff:
Thanks @bug-mirror , You did it again!

PR #233, opened 2017-10-20
merge strings for train 33708

bug-mirror:


shane-tomlinson:
Closing, not needed until next week.

PR #246, opened 2018-01-19
merge strings for train 19030

bug-mirror:


vbudhram:
from mtg: @vbudhram to follow up on how/when this gets merged.

vbudhram:
The last few of these types of PR's have been closed/merged, but there doesn't seem to be a pattern to it. Will get with vlad to get more information about this process.

PR #252, opened 2018-03-02
merge strings for train 23218

bug-mirror:


vladikoff:
Too early, @ryanfeeley doesn't have the final strings for this yet

PR #255, opened 2018-03-09
merge strings for train 32024

bug-mirror:


vladikoff:
Not today @bug-mirror !

PR #256, opened 2018-03-16
merge strings for train 11221

bug-mirror:


vladikoff:
Done in https://github.com/mozilla/fxa-content-server-l10n/pull/257

TonnesM:
"Scan the QR code in your app and confirm it is setup correctly …"

Shouldn’t that be "set up"?

Additionally, localizers may wonder what "it" applies to (either the two-factor authentication, the app, or even the QR code).

vladikoff:
cc @ryanfeeley @vbudhram

ryanfeeley:
"set up" is correct.

PR #279, opened 2018-08-08
fix(pa_IN): Fix the broken links in pa_IN

shane-tomlinson:
@mozilla/fxa-devs - r?

### DIFF
--- a/locale/pa_IN/LC_MESSAGES/server.po
+++ b/locale/pa_IN/LC_MESSAGES/server.po
@@ -1,4 +1,4 @@
-# 
+#
 msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
> How did you notice these btw? Do we have automated tests for different locales that caused them to show up, or did we just luck out that you happened to spot them?

We have an l10n linter that @TDA wrote a few years back. It checks whether there are any malformed or unexpected HTML tags/attributes. The linter runs on travis for every checkin and sends a few of us an email when they start failing. Want me to add you to the list?

shane-tomlinson:
Here's a link to the failing travis test BTW: https://travis-ci.org/mozilla/fxa-content-server-l10n/builds/413517550?utm_source=email&utm_medium=notification

philbooth:
> Want me to add you to the list?

Yes please!

PR #290, opened 2018-09-14
merge strings for train 17600

bug-mirror:


shane-tomlinson:
Closing in favor of #291

PR #294, opened 2018-10-05
merge strings for train 22335

bug-mirror:


shane-tomlinson:
Not needed this week, closing.

PR #298, opened 2018-11-02
merge strings for train 46438

bug-mirror:


shane-tomlinson:
Not needed until Nov 9th. Closing.

PR #307, opened 2018-11-19
Use Mozilla plural forms in AR locale

vladikoff:
Fixes https://github.com/mozilla/fxa-auth-server/issues/2714

@flodolo @mathjazz Any ideas here why the current Plural form would be causing problems?

If I change it to this one or switch it with a Russian one, then it fixes https://github.com/mozilla/fxa-auth-server/issues/2714 and Arabic translations start working again. Have you seen that before?

Ref: https://www.arabeyes.org/Plural_Forms

cc @philbooth

### DIFF
--- a/locale/ar/LC_MESSAGES/client.po
+++ b/locale/ar/LC_MESSAGES/client.po
@@ -11,7 +11,7 @@ msgstr ""
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"Plural-Forms: nplurals=6; plural=(n==0 ? 0 : n==1 ? 1 : n==2 ? 2 : n%100>=3 && n%100<=10 ? 3 : n%100>=11 ? 4 : 5);\n"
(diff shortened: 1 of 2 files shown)


vladikoff:
Here are the generated structures for template from PO files:

Arabic: 

https://gist.githubusercontent.com/vladikoff/becda0d0ccb308444071fee4f9cd2eaf/raw/d3077e653adda7152fd5b319813c73d00ca9cbec/ar-gen

Russian:

https://gist.githubusercontent.com/vladikoff/7c240cca29b3113839b52a9d91529719/raw/fb0c5bbbdf1b0a0fabe325c70fd4f54be68bbf65/ru-gen

flodolo:
I guess this would require changes in Pontoon's configuration, otherwise it will get reset at the first commit. Having said that, I don't think that swapping forms like this is OK, because it would mess up translations for any Gettext proejct. And I'm not aware of similar issues, e.g. in [AMO](https://github.com/mozilla/addons-frontend/blob/master/locale/ar/LC_MESSAGES/amo.po#L14-L15).

Also note that I'm not in charge of localization for this project, @peiying2 is.

mathjazz:
Agreed with @flodolo: Changing the plural rule would likely break other projects.

PR #334, opened 2019-04-05
merge strings for train 44772

bug-mirror:


shane-tomlinson:
Not last week, next week!

PR #335, opened 2019-04-08
fix(l10n): fix french translation for post change primary email

vbudhram:
@shane-tomlinson @vladikoff Never fixed an l10n issue before so not sure if this is the right process?

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1542704


vladikoff:
Thanks, that's so strange. why would anyone change that

PR #338, opened 2019-04-26
merge strings for train 13045

bug-mirror:


shane-tomlinson:
This PR doesn't contain any new strings, the Lambda function will need to be updated to work with the monorepo. Closing.

shane-tomlinson:
Replaced by #339

PR #340, opened 2019-04-26
Fix string extraction script

vladikoff:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1547332

### DIFF
--- a/scripts/extract-and-pull-request.sh
+++ b/scripts/extract-and-pull-request.sh
@@ -5,12 +5,11 @@ set -e
 rm -rf workspace
 mkdir workspace
 cd workspace
-git clone https://github.com/mozilla/fxa-auth-server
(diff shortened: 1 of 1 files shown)


scripts/extract-and-pull-request.sh
```diff
@@ -5,12 +5,12 @@ set -e
 rm -rf workspace
 mkdir workspace
 cd workspace
-git clone https://github.com/mozilla/fxa-auth-server
-cd fxa-auth-server && npm i && cd ..
+git clone https://github.com/mozilla/fxa
+cd fxa/packages/fxa-auth-server && npm ci && cd ../../..
 git clone https://github.com/mozilla/fxa-content-server
```
philbooth: This line should be deleted too right?

PR #365, opened 2019-07-31
added tabindex to legal links

dannycoates:
goes with https://github.com/mozilla/fxa/pull/2014

I _think_ this should get merged *before* the strings next get extracted and merged. Not sure, maybe it doesn't matter.


shane-tomlinson:
> I think this should get merged before the strings next get extracted and merged. Not sure, maybe it doesn't matter.

The tests are failing because the tabindex is unexpected, https://github.com/mozilla/fxa/pull/2014 needs to merge first so the tests pass.

dannycoates:
@shane-tomlinson r?

PR #375, opened 2019-09-30
merge strings for train 147

shane-tomlinson:


scripts/merge_po.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -ex
```
shane-tomlinson: @vladikoff - I noticed a problem after merging train-147 strings and ran `grunt l10n-lint` locally, it reported loads of failures. After digging in a bit, I saw that the failing locales contained duplicate strings, but these weren't caught by the script because we kept processing if msgmerge failed.

vladikoff: yeah makes sense!

PR #399, opened 2019-12-20
merge strings for train 22223

bug-mirror:


clouserw:
merged a newer version already

PR #404, opened 2020-02-28
merge strings for train 1000

clouserw:


clouserw:
I had to make this manually as the cron job failed.  Also added a bonus of adding `msguniq` to our merge script so we can stop having conflicts

PR #406, opened 2020-03-06
merge strings for train 2318

bug-mirror:


peiying2:
Changes in `en` is good to go from my side, but conflicts need to be resolved before merging.

PR #417, opened 2020-05-22
fix l10n extraction to use yarn

clouserw:


clouserw:
This fixes it for me locally.  The s/sh/bash/ change is because I don't think sh has a $RANDOM.  Maybe circleci is aliasing it for us?

dannycoates:
there was one more thing so i made #418

PR #420, opened 2020-05-29
merge strings for train 23086

bug-mirror:


locale/templates/LC_MESSAGES/client.pot
```diff
@@ -112,77 +119,77 @@ msgstr ""
...
 "possible."
 msgstr ""
 
-#: .es5/views/support.js:226
+#: .es5/views/support.js:237
 msgid ""
 "Thank you for reaching out to Mozilla Support about "
-"<b>%(escapedLowercaseTopic)s</b> for <b>%(escapedSelectedPlan)s</b>. We'll "
-"contact you via email as soon as possible."
+"<b>%(escapedLowercaseTopic)s</b> for <b>%(escapedSelectedProduct)s</b>. "
```
flodolo, line 141: I got pinged by the Italian localizers, asking how to deal with this.

I might be wrong, but you're taking original translations starting with  uppercase, and reusing them after making them lowercase via JS.
https://github.com/mozilla/fxa/blob/07cefc18148d0b17f5240120a84fe151a5032088/packages/fxa-content-server/app/scripts/models/support-form.js#L15

That is incredibly fragile (I'd go as far as to say it's broken):
- JS might not be able to manage the correct transformation for locales.
- You don't know what capitalization a language should use in the middle of a sentence.
- In the middle of a sentence it might need a different case.

peiying2: I didn't wasn't aware of this part, only looking for the en-US section. And if I did, I would have not understood this part. Thanks for pointing it out. And also, I need to check `template` section too.

flodolo: You don't need to check both, check either `en-US` or `templates`.

As for the issue, to be honest it's really hard to catch (assuming it's confirmed).

PR #421, opened 2020-06-02
feat: Add payments server to our L10n scripts

clouserw:


clouserw:
I ran this locally and it copied the file to the right place

scripts/extract_strings.sh
```diff
@@ -77,6 +84,8 @@ cp -r $CONTENT_DIR/locale/templates/* $L10N_DIR/locale/templates
 msgfilter -i $L10N_DIR/locale/sr/LC_MESSAGES/client.po -o $L10N_DIR/locale/sr_Latn/LC_MESSAGES/client.po recode-sr-latin
 msgfilter -i $L10N_DIR/locale/sr/LC_MESSAGES/server.po -o $L10N_DIR/locale/sr_Latn/LC_MESSAGES/server.po recode-sr-latin
 
+cp $PAYMENTS_DIR/public/locales/en-US/*.ftl $L10N_DIR/locale/templates
```
clouserw, line 87: I guess one note here is that it's copying *.ftl but only main.ftl is in the repository.  So, when we add new .ftl files it will copy them over but they don't get `git add`'d

PR #432, opened 2020-07-14
Add a few new locale clones.

clouserw:


clouserw:
We decided on doing this a different way.

Issue #473, opened 2021-02-22
Inconsistent ending periods for the "There was a problem" strings


ItielMaN:
```
Sorry, there was a problem refreshing the list of connected services.
There was a problem replacing your recovery codes.
There was a problem uploading your profile picture
There was a problem deleting your avatar
There was a problem updating your display name.
There was a problem creating this email.
There was a problem sending the verification code.
There was a problem retrieving your code.
There was a problem verifying your recovery code.
Sorry, there was a problem refreshing the recovery key.
Sorry, there was a problem verifying your session
Sorry, there was a problem refreshing that email.
Sorry, there was a problem re-sending the verification code.
Sorry, there was a problem changing your primary email.
Sorry, there was a problem deleting this email.
Sorry, there was a problem verifying your session.
Sorry, there was a problem refreshing two-step authentication.
Sorry, there was a problem verifying your session
```

Some of these have ending periods, some don't.

peiying2:
@6a68

Issue #470, opened 2021-02-19
Firefox Accounts incorrectly used for Firefox Account in context


peiying2:
firefox-accounts/settings.ftl

Problem: 
- cs-disconnect-lost-advice-content = Since your device was lost or stolen, to keep your information safe, you should change your { -product-firefox-accounts } password in your account settings. You should also look for information from your device manufacturer about erasing your data remotely.

- delete-account-confirm-title = You’ve connected your { -product-firefox-accounts } to { -brand-mozilla } products that keep you secure and productive on the web:

Should `{ -product-firefox-accounts }` be `{ -product-firefox-account }` (singular)? 

When submit the fixes, the string IDs need to be different. We can add "rev" or "2" at the end of the current IDs to differentiate.

peiying2:
@6a68 or @chenba

PR #475, opened 2021-02-23
Pull in settings updates

6a68:
* Fixes for bugs #470 and #473

* Sentence case adjustments to match the designs


peiying2 (approved):
Looks good.

locale/templates/settings.ftl
```diff
@@ -247,19 +250,24 @@ delete-account-password-input =
  .label = Enter password
 
 delete-account-cancel-button = Cancel
-delete-account-delete-button = Delete Account
+delete-account-delete-button = Delete
```
peiying2, line 253: @6a68 oops, I overlooked this one. This needs to have a new string ID. Can we fix it ASAP?

PR #476, opened 2021-02-24
Update a stray stale l10n ID

6a68:
@peiying2 Here's a fix for that stale ID from #475


peiying2 (approved):
Thank you for  the fix.

PR #477, opened 2021-02-26
merge strings for train 35171

bug-mirror:


locale/templates/main.ftl
```diff
@@ -76,8 +77,8 @@ product-plan-not-found = Plan not found
 product-no-such-plan = No such plan for this product.
 
 ## payment legal blurb
-payment-legal-copy-stripe-paypal = { -brand-name-mozilla } uses Stripe and Paypal for secure payment processing.
-payment-legal-link-stripe-paypal = View the <stripePrivacyLink>Stripe privacy policy</stripePrivacyLink> and <paypalPrivacyLink>Paypal privacy policy</paypalPrivacyLink>.
+payment-legal-copy-stripe-paypal = { -brand-name-mozilla } uses Stripe and { -brand-name-paypal } for secure payment processing.
```
peiying2, line 80: Need a new string ID or the changes won't notify to those already localized the strings.

peiying2: @chenba This PR was older than the one you merged. This one needs to be fixed. It might be easier to start a new one because of conflicts you will run into.

chenba: I'll file an issue to update the id in the FxA repo.  As for the PR, not sure if I understand the process correctly, but wouldn't the new PR override the older one(s)?  That is, whichever PR that's from the most recent Friday extraction run would have all the latest changes?

peiying2: I can close this PR. I think the one you merged should have picked up the l10n updates from this PR. Will wait for the fix in the next push.

PR #503, opened 2021-06-24
merge strings for train 26628

bug-mirror:


locale/templates/settings.ftl
```diff
@@ -539,6 +539,7 @@ tfa-row-change-modal-explain = You won’t be able to undo this action.
 auth-error-102 = Unknown account
 auth-error-103 = Incorrect password
 auth-error-110 = Invalid token
+auth-error-114 = You've tried too many times. Please try again { $retryAfter }.
```
peiying2, line 542: Please add comments for what { $retryAfter} stands for

peiying2, line 542: Need to add a variable at the beginning of this block what { $retryAfter } is for

peiying2 (changes requested):
Need to explain what the variable stands for

PR #504, opened 2021-07-01
merge strings for train 17776

bug-mirror:


peiying2 (changes requested):
Please provide context to how the  variable is used
Use curly quote (single or double) in .ftl files.

PR #516, opened 2021-09-08
retrieve ftl files in fxa-auth-server

jodyheavener:
We're transitioning the `fxa-auth-server` to an updated email system, part of which includes using Fluent.

This PR (hopefully) updates the scripts to look for FTL files in the auth-server.


LZoog (approved):
LGTM

PR #519, opened 2021-09-22
merge strings for train 47665

bug-mirror:


jodyheavener:
The script has been broken for a bit, so this is the first run in what looks like about 20 days. Please let me know if anything looks off here.

This is also the first run introducing new FxA Emails that use Fluent. The new file is `locale/templates/auth.ftl`.

peiying2 (approved):

PR #527, opened 2021-10-15
Add auth.ftl template

LZoog:
Putting this up as a draft PR for an initial review.


locale/templates/auth.ftl
```diff
@@ -0,0 +1,305 @@
...
+
+## Email content
+
+fxa-privacy-url = { -brand-mozilla } Privacy Policy
+fxa-service-url = { -brand-firefox } Cloud Terms of Service
+
+subplat-automated-email = This is an automated email; if you received it in error, no action is required.
+subplat-privacy-plaintext = Privacy notice:
+
+subplat-terms-policy-plaintext = { subplat-terms-policy }:
```
flodolo, line 30: Could we have these put closer to the referenced string? e.g.

```
subplat-terms-policy = Terms and cancellation policy
subplat-terms-policy-plaintext = { subplat-terms-policy }:
```

locale/templates/auth.ftl
```diff
@@ -0,0 +1,305 @@
...
+
+# The user has a low number of valid recovery codes remaining for use
+codes-reminder-title = Low recovery codes remaining
+codes-reminder-description = We noticed that you are running low on recovery codes. Please consider generating new codes to avoid getting locked out of your account.
+codes-generate = Generate codes
+codes-generate-plaintext = { codes-generate }:
+lowRecoveryCodes-subject =
+    { NUMBER($numberRemaining) ->
+        [one] 1 recovery code remaining
+       *[other] { NUMBER($numberRemaining) } recovery codes remaining
```
flodolo, line 106: You should not need to use `NUMBER()` here, or at the beginning

```
lowRecoveryCodes-subject =
    { $numberRemaining ->
        [one] 1 recovery code remaining
       *[other] $numberRemaining recovery codes remaining
   }
```

locale/templates/auth.ftl
```diff
@@ -0,0 +1,305 @@
...
+    }
+lowRecoveryCodes-action = { codes-generate }
```
flodolo, line 108: For these, you should just put the text, not reference another message
https://firefox-source-docs.mozilla.org/l10n/fluent/review.html#message-references

It makes sense in the HTML vs plain strings, because you want to enforce consistency, but here it just makes localization more opaque, and defies the point of having translation memory in tools like Pontoon.

locale/templates/auth.ftl
```diff
@@ -0,0 +1,305 @@
...
+
+# Variables:
+#  $clientName (String) - A client the user hasn't signed into before (e.g. Firefox, Sync)
+newDeviceLogin-subject = New sign-in to { $clientName }
+newDeviceLogin-title = { newDeviceLogin-subject }
+newDeviceLogin-action = { manage-account }
```
flodolo, line 114: Same for these strings, it should just be 

```
newDeviceLogin-action = Manage account
```

locale/templates/auth.ftl
```diff
@@ -0,0 +1,305 @@
...
+verificationReminderSecond-subject = Final reminder: Activate your account
+verificationReminderSecond-title = Still there?
+verificationReminderSecond-description = Almost a week ago you created a { -brand-firefox } Account but never verified it. We’re worried about you.
+verificationReminderSecond-sub-description = Confirm this email address to activate your account and let us know you’re okay.
+verificationReminderSecond-action = { confirm-email }
+
+verify-title = Activate the { -brand-firefox } family of products
+verify-description-common = Confirm your account and get the most out of { -brand-firefox } everywhere you sign in
+verify-description-plaintext = { verify-description-common }.
+verify-description = { verify-description-common } starting with:
```
flodolo, line 256: Never concatenate full sentences like this.

This should be two separate strings.

```
verify-description-plaintext = Confirm your account and get the most out of { -brand-firefox } everywhere you sign in.
verify-description = Confirm your account and get the most out of { -brand-firefox } everywhere you sign in starting with:
```

https://mozilla-l10n.github.io/documentation/localization/dev_best_practices.html#splitting-and-composing-sentences

locale/templates/auth.ftl
LZoog: When I try what you suggested, the string literal "$numberRemaining" is printed for `*[other]*`:

```
Subject: $numberRemaining recovery codes remaining
```

It appears to work fine with this which I've gone with:
```
lowRecoveryCodes-subject =
    { $numberRemaining ->
        [one] 1 recovery code remaining
       *[other] { $numberRemaining } recovery codes remaining
   }
```

LZoog:
@flodolo Thanks so much for your suggestions and patience here. I just pushed an update, I talked with some folks about the Accounts v account issue and we got that worked out so I went ahead and made that change and others you suggested.

I'm going to be writing some docs for the FxA/Subplat team by the end of the year around more info on the l10n process and what your team needs for Fluent so things you've shared here as well as the Slack channel shouldn't get lost on folks working on similar things in the future. 🤞

Let me know if anything else stands out in the template file, I hope to trash the previous migration script and get a new one going soon that we can reuse later. Early next year we're likely migrating the rest of FxA's content server over to Fluent so I'll try to make sure that goes much more smoothly.

LZoog:
I've been able to rework the script using existing parsers and pushed an update containing the translations. Since my `templates/auth.ftl` changes based on the feedback above may not have been reviewed yet, I kept them [in a separate commit](https://github.com/mozilla/fxa-content-server-l10n/pull/527/commits/b813c72fa476cf5c01dc138fbf63127a27f46193). See the full file [here](https://github.com/mozilla/fxa-content-server-l10n/blob/b813c72fa476cf5c01dc138fbf63127a27f46193/locale/templates/auth.ftl).

I checked with our legal team about the "Firefox Cloud" phrase which I noticed we're not translating at the moment, and I was told (at least for now) let's continue to treat it as a brand.

It's very helpful that we don't have to have comments in the localized files, but one thing to note is that I went ahead and used the translation for "Firefox Accounts" if it existed, though it is now _lowercase_ "Firefox accounts" across the board. I assume localizers will see the comment I left noting it needs to be lowercased and fix that if something is not right? I can change this to whatever is preferred if not. [Here's the commit](https://github.com/mozilla/fxa-content-server-l10n/pull/527/commits/f11059f948a1c418a9d846b35c510648649a2532) containing just the translations from [this script](https://github.com/LZoog/po-translations-to-ftl).

locale/en/auth.ftl
```diff
@@ -0,0 +1,314 @@
...
+codes-generate-plaintext = { codes-generate }:
+lowRecoveryCodes-action = Generate codes
+lowRecoveryCodes-subject =
+    { $numberRemaining ->
+        [one] 1 recovery code remaining
+       *[other] { $numberRemaining } recovery codes remaining
+   }
+
+# Variables:
+#  $clientName (String) - A client the user hasn't signed into before (e.g. Firefox, Sync)
```
flodolo, line 111: This should be a group comment, since it applies to more than one string.

The whole bunch of lines would look like this.

```
## Variables:
##  $clientName (String) - A client the user hasn't signed into before (e.g. Firefox, Sync)

newDeviceLogin-subject = New sign-in to { $clientName }
newDeviceLogin-title = New sign-in to { $clientName }

##

newDeviceLogin-action = Manage account
```

locale/en/auth.ftl
```diff
@@ -0,0 +1,314 @@
...
+passwordResetAccountRecovery-title = Your account password was reset with a recovery key
+passwordResetAccountRecovery-description = You have successfully reset your password using a recovery key from the following device:
+passwordResetAccountRecovery-action = Create new recovery key
+passwordResetAccountRecovery-regen-required = You will need to generate a new recovery key.
+# After the colon, there's a link to https://accounts.firefox.com/settings/account_recovery
+passwordResetAccountRecovery-create-key = Create new recovery key:
+
+postAddAccountRecovery-subject = Account recovery key generated
+postAddAccountRecovery-title = Account recovery key generated
+postAddAccountRecovery-description = You have successfully generated an account recovery key for your { -brand-firefox } Account using the following device:
```
flodolo, line 143: I assume we want to fix all existing instances of brand-firefox + account

```suggestion
postAddAccountRecovery-description = You have successfully generated an account recovery key for your { -product-firefox-account } using the following device:
```

locale/en/auth.ftl
```diff
@@ -0,0 +1,314 @@
...
+#  $invoiceTotal (String) - The amount of the subscription invoice, including currency, e.g. $10.00
+#  $nextInvoiceDateOnly (String) - The date of the next invoice, e.g. 2016/01/20
+subscriptionReactivation-content = Your billing cycle and payment will remain the same. Your next charge will be { $invoiceTotal } on { $nextInvoiceDateOnly }. Your subscription will automatically renew each billing period unless you choose to cancel.
+
+subscriptionsPaymentExpired-subject = Credit card for your subscriptions is expiring soon
+subscriptionsPaymentExpired-title = Your credit card is about to expire
+subscriptionsPaymentExpired-content = The credit card you’re using to make payments for the following subscriptions is about to expire.
+# Variables:
+#  $productName (String) - The name of the subscribed product, e.g. Mozilla VPN
+subscriptionsPaymentExpired-name = { $productName }
```
flodolo, line 238: Do we actually need this exposed as an FTL message, since it's just the product name?

flodolo:
> I assume localizers will see the comment I left noting it needs to be lowercased and fix that if something is not right?

We need to make it's communicated to localizers (@peiying2).

Left a couple of comments, but this looks great 👍🏼

locale/en/auth.ftl
LZoog: Oh yes, this was missed, thanks for the catch.

locale/en/auth.ftl
LZoog: I don't think so, will remove 👍

LZoog:
@flodolo @peiying2 I pushed an update addressing the last couple of comments.

I also took some time to manually verify a few more of these locale FTL files against the translations from their po counterpart, and they're looking good to me. I know it would be nice to have some test fixtures to compare against... but we have already spent a considerable amount of time on this and I've asked our QA team not to to test our new emails in staging until this is in. I'm feeling pretty confident in where this is and know the parsers themselves have pretty extensive tests, and because I need to balance priorities I would recommend we skip that for now and to revisit it later, but feel free to let me know your thoughts.

flodolo (approved):
I did a quick check of Italian, and it seems OK.

peiying2:
I spot-checked zh_CN and zh_TW versions, and specifically looked for the ones containing `{ -product-firefox-account }` in Italian version but I couldn't find them in these two, following the same string ID. Then I checked other languages, they have the same issue. Note, I didn't just look for the brand name placeables but I went by the string IDs.

flodolo:
It wouldn't be possible to replace localized variations of "Firefox account" with the term name. Those will show up as warnings in Pontoon, and localizers will need to update them (or leave them as is). That should be part of the outreach to community.

PR #535, opened 2021-11-18
merge strings for train 36803

bug-mirror:


peiying2 (approved):

locale/templates/settings.ftl
```diff
@@ -479,7 +479,7 @@ row-defaults-status = None
 
 rk-header = Recovery key
 rk-enabled = Enabled
-rk-not-set = Not set
+rk-not-set = Not Set
```
peiying2, line 482: If the change is a must, then we need a new string ID otherwise, reusing the same ID will not trigger Pontoon as a new or revised string - this is the case for those locales already translated the strings, including en-CA and en-GB.

locale/templates/settings.ftl
```diff
@@ -535,7 +535,7 @@ se-content-note = Note: a secondary email won’t restore your information — y
 tfa-row-header = Two-step authentication
 tfa-row-disabled = Two-step authentication disabled.
 tfa-row-enabled = Enabled
-tfa-row-not-set = Not set
+tfa-row-not-set = Not Set
```
peiying2, line 538: If the change is a must, then we need a new string ID otherwise, reusing the same ID will not trigger Pontoon as a new or revised string - this is the case for those locales already translated the strings, including en-CA and en-GB.

peiying2 (commented):
@vbudhram comment on change of string using the same string IDs.

locale/templates/settings.ftl
dannycoates: we don't want to trigger any changes in pontoon for this. i made it so that our local env will match production for en locales

PR #538, opened 2021-12-02
merge strings for train 46704

bug-mirror:


locale/templates/settings.ftl
```diff
@@ -527,6 +535,8 @@ se-make-primary = Make primary
 se-default-content = Access your account if you can’t log in to your primary email.
 se-content-note = Note: a secondary email won’t restore your information — you’ll
   need a <a>recovery key</a> for that.
+# Default value for the secondary email
+se-secondary-email-none = None
 
 ##
```
peiying2, line 541: Remove this if not used.

locale/templates/settings.ftl
```diff
@@ -214,6 +216,10 @@ mvs-enter-verification-code-desc = Please enter the verification code that was s
 msv-cancel-button = Cancel
 msv-submit-button = Verify
 
+## Auth-server based errors that originate from backend service
```
peiying2, line 219: Is  this comment for a single string or a section?

locale/templates/settings.ftl
```diff
@@ -364,7 +370,9 @@ add-secondary-email-enter-address =
 add-secondary-email-cancel-button = Cancel
 add-secondary-email-save-button = Save
 
-##
+## Auth-server based errors that originate from backend service
```
peiying2, line 373: Is this comment for a single string or for a section?

locale/templates/settings.ftl
```diff
@@ -479,7 +487,7 @@ row-defaults-status = None
 
 rk-header = Recovery key
 rk-enabled = Enabled
-rk-not-set = Not set
+rk-not-set = Not Set
```
peiying2, line 490: Why is this capitalized? The change won't impact l10n if the string is localized.

locale/templates/settings.ftl
```diff
@@ -535,7 +545,7 @@ se-content-note = Note: a secondary email won’t restore your information — y
 tfa-row-header = Two-step authentication
 tfa-row-disabled = Two-step authentication disabled.
 tfa-row-enabled = Enabled
-tfa-row-not-set = Not set
+tfa-row-not-set = Not Set
```
peiying2, line 548: Why is this capitalized? The change won't impact l10n if the string is localized.

peiying2 (commented):
See in-line comments

locale/templates/settings.ftl
StaberindeZA: Hi @peiying2 this string should have been in the same section as the other `auth-error-xxx` strings. 

This seems to be an issue in the code base, I'll make the necessary updates so that this duplicate section is removed and the string is is added to the other `auth-error-xxx` strings.

locale/templates/settings.ftl
StaberindeZA: Hi @peiying2 this string should have been in the same section as the other `auth-error-xxx` strings. 

This seems to be an issue in the code base, I'll make the necessary updates so that this duplicate section is removed and the string is is added to the other `auth-error-xxx` strings.

locale/templates/settings.ftl
clouserw: This is used @ https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx#L139

locale/templates/settings.ftl
clouserw: Thanks for the fix

locale/templates/settings.ftl
clouserw: As mentioned in a previous PR, this is just for our tests.  No change to L10n necessary

PR #539, opened 2021-12-08
Transfer existing translations for newly updated emails into auth.ftl

LZoog:
We've created a few more email templates using the new stack with Fluent and want to transfer translations over along with the `templates/auth.ftl` update.


locale/en/auth.ftl
```diff
@@ -215,6 +218,16 @@ recovery-title = Need to reset your password?
...
 
+#  Variables:
+#  $productName (String) - The name of the subscribed product, e.g. Mozilla VPN
+subscriptionAccountDeletion-subject = Your { $productName } subscription has been cancelled
+subscriptionAccountDeletion-title = Sorry to see you go
+#  Variables:
+#  $productName (String) - The name of the subscribed product, e.g. Mozilla VPN
+#  $invoiceTotal (String) - The amount of the subscription invoice, including currency, e.g. $10.00
+#  $invoiceDateOnly (String) - The date of the next invoice, e.g. 01/20/2016
+subscriptionAccountDeletion-content-cancelled = You recently deleted your Firefox Account. As a result, we've cancelled your { $productName } subscription. Your final payment of { $invoiceTotal } was paid on { $invoiceDateOnly }.
```
LZoog, line 229: Just noticed `we've` needs a curly quote. Will fix.

locale/templates/auth.ftl
```diff
@@ -215,6 +218,16 @@ recovery-title = Need to reset your password?
...
 
+#  Variables:
+#  $productName (String) - The name of the subscribed product, e.g. Mozilla VPN
+subscriptionAccountDeletion-subject = Your { $productName } subscription has been cancelled
+subscriptionAccountDeletion-title = Sorry to see you go
+#  Variables:
+#  $productName (String) - The name of the subscribed product, e.g. Mozilla VPN
+#  $invoiceTotal (String) - The amount of the subscription invoice, including currency, e.g. $10.00
+#  $invoiceDateOnly (String) - The date of the next invoice, e.g. 01/20/2016
+subscriptionAccountDeletion-content-cancelled = You recently deleted your Firefox Account. As a result, we've cancelled your { $productName } subscription. Your final payment of { $invoiceTotal } was paid on { $invoiceDateOnly }.
```
LZoog, line 229: Just noticed `we've` needs a curly quote. Will fix.

peiying2: `Firefox Account` needs to be in a placeable

peiying2 (changes requested):
Need to update Firefox Account with a placeable.

LZoog:
@peiying2 updated. 🙂

peiying2 (commented):
@LZoog Thanks for the fix. The smart quote or apostrophe somehow looks like an "opening" single quote, but should be a "closing" one (direction different). Take a look of the French content (the whole file, not just the new content). Am I just noticing this or has it been always like this?

LZoog:
@peiying2 Hmm, it looks like a "closing" quote to me. Is there one in particular that you see? For example...

From [this existing translation](https://github.com/mozilla/fxa-content-server-l10n/blob/master/locale/fr/LC_MESSAGES/server.po#L117), we can see:

```
Ceci est un message automatique ; si vous l’avez reçu par erreur, vous n’avez rien à faire.
```

Then [here](https://github.com/mozilla/fxa-content-server-l10n/blob/master/locale/fr/auth.ftl#L25) in the leveraged translation, we also see:
```
Ceci est un message automatique ; si vous l’avez reçu par erreur, vous n’avez rien à faire.
```

peiying2:
> @peiying2 Hmm, it looks like a "closing" single quote to me. Is there one in particular that you see? For example...
> 
> From [this existing translation](https://github.com/mozilla/fxa-content-server-l10n/blob/master/locale/fr/LC_MESSAGES/server.po#L117), we can see:
> 
> ```
> Ceci est un message automatique ; si vous l’avez reçu par erreur, vous n’avez rien à faire.
> ```
> 
> Then [here](https://github.com/mozilla/fxa-content-server-l10n/blob/master/locale/fr/auth.ftl#L25) in the leveraged translation, we also see:
> 
> ```
> Ceci est un message automatique ; si vous l’avez reçu par erreur, vous n’avez rien à faire.
> ```
> 
> Edit: I might've misunderstood, we just caught that the `we've`s that we fixed are using an opening single quote and should be using a closing quote but this only affects the English version. Will push again shortly but it won't affect the locales.
> 
> Edit2: Pushed the update to fix those. 🙂

Thank you for the fixes!! I noticed `cak` have weird apostrophe going in both directions plus straight quote. We will leave them to the community to fix it.

peiying2 (approved):
Thank you for the fixes. LGTM!

PR #540, opened 2021-12-09
merge strings for train 54488

bug-mirror:


locale/templates/main.ftl
```diff
@@ -413,3 +411,8 @@ new-user-submit = Subscribe Now
 
 manage-pocket-title = Looking for your { -brand-name-pocket } premium subscription?
 manage-pocket-body = To manage it, <a>click here</a>.
+
+payment-method-header = Choose your payment method
+# $prefix (string) - If header is part of a multi step process and needs a header. eg. '2.'
+payment-method-header-prefix = { $prefix } Choose your payment method
```
flodolo: This looks like hard-coded content. How do you know which type of numeral, spacing, etc. should go there based on locale?

Also, why does `auth.ftl` exist in `en-US` too, but the other two .ftl files don't?

clouserw: > This looks like hard-coded content. How do you know which type of numeral, spacing, etc. should go there based on locale?
This looks like it came from https://github.com/mozilla/fxa/pull/11006 .  @StaberindeZA @chenba - can you answer the question here?

> Also, why does auth.ftl exist in en-US too, but the other two .ftl files don't?
I'm assuming it's a mistake an unused, but @LZoog might know

chenba: > This looks like hard-coded content. How do you know which type of numeral, spacing, etc. should go there based on locale?
> 

Maybe I can explain how the message is used, and @flodolo can make a suggestion on what to do?  I don't think I have the answer.

The message is used in two places in the code; one with a value passed for the `$prefix` (hard-coded to "2.") and one without.  In the case with the `$prefix`, it is the second step to the message with the id `new-user-step-1`.  That message has "1." hard-coded into the message (https://github.com/mozilla/fxa/blob/main/packages/fxa-payments-server/public/locales/en-US/main.ftl#L396).

Would it be better if we define a message without the step prefix, and then define another for the prefix'd message, e.g., `new-user-step-2 = 2. {payment-method-heading}`?

LZoog: > > Also, why does auth.ftl exist in en-US too, but the other two .ftl files don't?
> 
> I'm assuming it's a mistake and unused, but @LZoog might know

This is a mistake. I'll get a PR up removing these tomorrow AM. 👍

flodolo: > Would it be better if we define a message without the step prefix, and then define another for the prefix'd message, e.g., `new-user-step-2 = 2. {payment-method-heading}`?

That sounds like a good plan. I would only avoid `-2` in the ID, because that's typically use to version a string when you need to change the text.

LZoog: @flodolo I'm guessing `master` isn't protected from pushing directly due to Pontoon's sync. I removed `en_US/auth.ftl` as well as `en/auth.ftl` since we don't have `main.ftl` or `settings.ftl` in the `en` locale either, but I thought I'd already branched off of `master` and I ran a `git push` which pushed directly to `master`.

Just for awareness, [this commit](https://github.com/mozilla/fxa-content-server-l10n/commit/ea811ec26d122bef3358bbe629b37860473357c5) was meant to be in a PR first, but I sent this to @peiying2 right after I pushed for confirmation, and she says it's fine, so I'll leave it as-is unless I'm told otherwise. Apologies for the direct push.

PR #541, opened 2021-12-16
merge strings for train 36061

bug-mirror:


locale/templates/auth.ftl
```diff
@@ -280,6 +344,35 @@ subscriptionsPaymentExpired-subject = Credit card for your subscriptions is expi
 subscriptionsPaymentExpired-title = Your credit card is about to expire
 subscriptionsPaymentExpired-content = The credit card you’re using to make payments for the following subscriptions is about to expire.
 
+subscriptionsPaymentProviderCancelled-subject = Payment information update required for Mozilla subscriptions
```
peiying2, line 347: Mozilla is a brand

locale/templates/auth.ftl
```diff
@@ -20,6 +20,9 @@
 # "Firefox Cloud" should be treated as a brand.
 -product-firefox-cloud = Firefox Cloud
 
+# Other brands
+-paypal = PayPal
```
peiying2, line 24: should consider change brand names to `-brand-paypal` to differentiate brand/product names from others.

peiying2 (changes requested):
Other than what's called on at string level, one general comment is on the `string level` comment - lots of repetitions. Should consider "section comment" as discussed here in this [doc](https://github.com/mozilla-l10n/documentation/blob/main/src/localization/dev_best_practices.md#fluent-files) . One example:  # `$productNameNew (String) - The name of the new subscribed product, e.g. Mozilla VPN`

xlisachan:
Thanks for feedback, @peiying2! We'll make the fix for the Mozilla and PayPal brand names now.

Regarding the section comment, can you confirm that it should look as follows?
`## Common email variables`
`# $productNameNew (String) - The name of the new subscribed product, e.g. Mozilla VPN`

Also, would it be preferable to have one section comment of all email variables within `locale/templates/auth.ftl` (where we will revise the PayPal brand name), or have each email have its own section comment of only the variables used in its `en.ftl`?

Please let us know when you have a moment. Thank you!

peiying2:
> Thanks for feedback, @peiying2! We'll make the fix for the Mozilla and PayPal brand names now.
> 
> Regarding the section comment, can you confirm that it should look as follows? `## Common email variables` `# $productNameNew (String) - The name of the new subscribed product, e.g. Mozilla VPN`
> 
> Also, would it be preferable to have one section comment of all email variables within `locale/templates/auth.ftl` (where we will revise the PayPal brand name), or have each email have its own section comment of only the variables used in its `en.ftl`?
> 
> Please let us know when you have a moment. Thank you!

Thank you for the fixes. Let's leave the "common variables" as they are. They are not a blocker for merging this PR. We can discuss how to best address common variables at section (feature) level.

LZoog:
@peiying2 I pushed an update to bug-mirror's PR with Lisa's fixes. 🙂

PR #542, opened 2021-12-23
merge strings for train 32993

bug-mirror:


locale/templates/main.ftl
```diff
@@ -344,6 +344,17 @@ plan-details-header = Product details
 plan-details-show-button = Show details
 plan-details-hide-button = Hide details
 plan-details-total-label = Total
+plan-details-list-price = List Price
+
+## coupons
```
peiying2, line 349: If this is a section comment, it needs to have an extra line break between this and the strings below. 

See documentation on this; https://projectfluent.org/fluent/guide/comments.html

locale/templates/auth.ftl
```diff
@@ -249,6 +288,34 @@ subscriptionAccountDeletion-title = Sorry to see you go
...
+subscriptionAccountReminderFirst-content-select = Select “Create Password” to set up a new password and finish verifying your account.
+subscriptionAccountReminderFirst-action = Create Password
+subscriptionAccountReminderFirst-action-plaintext = { subscriptionAccountReminderFirst-action }:
+
+subscriptionAccountReminderSecond-subject = Final reminder: Setup your account
+subscriptionAccountReminderSecond-title = Welcome to { -brand-firefox }!
+subscriptionAccountReminderSecond-content-info = A few days ago you created a { -product-firefox-account } but never verified it. We hope you’ll finish setting up your account, so you can use your new subscription.
+subscriptionAccountReminderSecond-content-select = Select “Create Password” to set up a new password and finish verifying your account.
+subscriptionAccountReminderSecond-action = Create Password
+subscriptionAccountReminderSecond-action-plaintext = { subscriptionAccountReminderSecond-action }:
```
peiying2, line 317: Do we need to expose this string?

locale/templates/auth.ftl
```diff
@@ -249,6 +288,34 @@ subscriptionAccountDeletion-title = Sorry to see you go
...
+#  $productName (String) - The name of the subscribed product, e.g. Mozilla VPN
+subscriptionAccountFinishSetup-content-create = Next, you’ll create a Firefox account password and download { $productName }.
+subscriptionAccountFinishSetup-action = Create a password
+
+subscriptionAccountReminderFirst-subject = Reminder: Finish setting up your account
+subscriptionAccountReminderFirst-title = You can’t access your subscription yet
+subscriptionAccountReminderFirst-content-info = A few days ago you created a { -product-firefox-account } but never verified it. We hope you’ll finish setting up your account, so you can use your new subscription.
+subscriptionAccountReminderFirst-content-select = Select “Create Password” to set up a new password and finish verifying your account.
+subscriptionAccountReminderFirst-action = Create Password
+subscriptionAccountReminderFirst-action-plaintext = { subscriptionAccountReminderFirst-action }:
```
peiying2, line 310: Does this need to be exposed?

locale/templates/auth.ftl
```diff
@@ -46,6 +47,10 @@ subplat-cancel-plaintext = { subplat-cancel }:
 subplat-reactivate = Reactivate subscription
 subplat-reactivate-plaintext = { subplat-reactivate }:
 subplat-update-billing = Update billing information
+subplat-privacy-policy = { -brand-mozilla } Privacy Policy
+subplat-privacy-policy-plaintext = { subplat-privacy-policy }:
```
peiying2, line 51: Do we need to expose this string to l10n?

locale/templates/auth.ftl
```diff
@@ -46,6 +47,10 @@ subplat-cancel-plaintext = { subplat-cancel }:
...
+subplat-cloud-terms = { -product-firefox-cloud } Terms of Service
+subplat-cloud-terms-plaintext = { subplat-cloud-terms }:
```
peiying2, line 53: Same here?

peiying2 (changes requested):
One change request in the main.ftl file. 

Question for strings that contain no localizable content such as this one ` subscriptionAccountReminderFirst-action-plaintext = { subscriptionAccountReminderFirst-action }: `

locale/templates/auth.ftl
clouserw: @xlisachan @LZoog

locale/templates/auth.ftl
clouserw: @xlisachan @LZoog

locale/templates/main.ftl
clouserw: @IvoJP

clouserw:
Thanks for the review.  Most folks are out for the holiday weekend so we'll probably have to fix these and re-extract next week.

peiying2:
> Thanks for the review. Most folks are out for the holiday weekend so we'll probably have to fix these and re-extract next week.

Sounds good. I will be off next week. Feel free to merge it afterwards

locale/templates/auth.ftl
LZoog: These strings expose the same string as the non-plaintext version (and we want them to always be identical, which is why we're using a peaceable) but they also have a colon, so we do need them. This is the same for the other comments.

locale/templates/main.ftl
IvoJP: I'm not sure how to handle this request - there isnt an extra line for groups anywhere else in that file so adding it for this section would be inconsistent. I also couldn't find anything in the documentation that recommended/required it (i could be wrong) aside from the example employing the change requested here.

locale/templates/auth.ftl
peiying2: @LZoog you are right, it contains  a`:` in addition to the brand name, so they should stay.

PR #543, opened 2021-12-30
merge strings for train 18762

bug-mirror:


locale/templates/main.ftl
```diff
@@ -344,6 +344,17 @@ plan-details-header = Product details
 plan-details-show-button = Show details
 plan-details-hide-button = Hide details
 plan-details-total-label = Total
+plan-details-list-price = List Price
+
+## coupons
```
peiying2: @IvoJP, I would like to move the conversation here from https://github.com/mozilla/fxa-content-server-l10n/pull/542/files which I will close.

https://github.com/mozilla-l10n/documentation/blob/main/src/localization/dev_best_practices.md#fluent-files and https://projectfluent.org/fluent/guide/comments.html give you examples how the comments are used, without adding instruction. If you see the #comment, it's immediately above the localizable strings. Others, they have an extra line space.

PR #544, opened 2022-01-06
merge strings for train 34316

bug-mirror:


locale/templates/LC_MESSAGES/server.pot
```diff
@@ -808,10 +836,10 @@ msgstr ""
 #: lib/senders/templates/subscriptionUpgrade.txt:3
```
peiying2: New string ID needed. Can consider change the number in the string ID.  https://github.com/mozilla-l10n/documentation/blob/main/src/localization/making_string_changes.md#why-is-it-necessary-to-use-new-ids

locale/templates/main.ftl
```diff
@@ -344,6 +344,17 @@ plan-details-header = Product details
 plan-details-show-button = Show details
 plan-details-hide-button = Hide details
 plan-details-total-label = Total
+plan-details-list-price = List Price
+
+## coupons
```
peiying2: @IvoJP, Please add an extra line between Section comment and the strings. Check the examples provided here: https://github.com/mozilla-l10n/documentation/blob/main/src/localization/dev_best_practices.md#fluent-files

locale/templates/LC_MESSAGES/server.pot
```diff
@@ -538,10 +541,10 @@ msgstr ""
 #: lib/senders/templates/subscriptionDowngrade.txt:2
 msgid ""
```
peiying2: Should this have a new ID? Check this doc for the reason. https://github.com/mozilla-l10n/documentation/blob/main/src/localization/making_string_changes.md#why-is-it-necessary-to-use-new-ids

peiying2 (changes requested):
Make sure revised strings have a new ID, otherwise, updated content won't trigger new string in Pontoon for communities to translate. Updated English and previously localized string will be out of sync.

peiying2:
@clouserw please help tag the developers who are contributing to the codes.

One thing I want to call out, is that when a string is updated, the string needs to be updated. We are pretty strict on Fluent, but this is the case for all file format. Without a new ID, Pontoon won't pick up the change of ID to trigger those strings as "untranslated" as the string was already translated because the ID remained the same.

peiying2 (approved):
Approve this PR so localized content can be pushed to production. Some of the issues raised will be followed up through Jira ticket
- Audit .ftl files on how "comments" are used 
- String ID in gettext file type with updated strings

PR #549, opened 2022-01-28
fix(content-server-l10n): fix errors in German translations

### LINKED ISSUE #548: [email templates l10n]: fix errors in German translations
Errors in the `locale/de/auth.ftl` are causing the following issues:
- [11783](https://github.com/mozilla/fxa/issues/11783) - App store badge missing from postVerify template when browser's language is set to German
- [11784](https://github.com/mozilla/fxa/issues/11784) - Special char displayed near text in postAddAccountRecovery template HTML when German is selected
- [11785](https://github.com/mozilla/fxa/issues/11785) - Variables not replaced with the correct text in subscriptionDowngrade and subscriptionUpgrade templates (HTML and plaintext) - German

xlisachan:
Closes #548

----
Within `mozilla/fxa`:
- Closes [11783](https://github.com/mozilla/fxa/issues/11783)
- Closes [11784](https://github.com/mozilla/fxa/issues/11784) 
- Closes [11785](https://github.com/mozilla/fxa/issues/11785)

### DIFF
--- a/locale/de/auth.ftl
+++ b/locale/de/auth.ftl
@@ -64,7 +64,7 @@ subplat-privacy-website-plaintext = { subplat-privacy }:
 body-android-badge = <img data-l10n-name="google-play-badge" alt="{ $productName } bei { -google-play } herunterladen">
 # Variables:
 #  $productName (String) - The name of the product to be downloaded, e.g. Mozilla VPN, or Firefox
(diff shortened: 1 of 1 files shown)


peiying2 (approved):
Thank you for the fixes. I will send a reminder to the contributor to watch out for these variables that seem to be a repeat but not exactly, so not to copy from a localized string of a different source altogether.

PR #553, opened 2022-02-17
merge strings for train 17980

bug-mirror:


locale/templates/settings.ftl
```diff
@@ -208,6 +209,15 @@ input-password-hide-aria = Hide password from screen.
...
 
 
+## Linked Accounts section
+
+la-heading = Linked Accounts
+la-description = You have authorized access to the following accounts.
+la-unlink-button = Unlink
+la-unlink-account-button = Unlink
+la-unlink-heading = Unlink from third party account
+la-unlink-content = Are you sure you want to unlink your { -brand-google } account? Unlinking your account does not automatically sign you out of those services, to do that you will need to manually sign out from the Connected services section.
```
peiying2: The punctuation are not correct. Should read like this `.... Unlinking your account does not automatically sign you out of those services. To do that, you will need to manually sign out from the Connected services section.`  - 1). change comma to period for the first sentence. 2). Then start with `To do that,` (add a comma after).

peiying2 (changes requested):
Need to fix the punctuation in the commented string.

peiying2:
@LZoog  not sure who should look into this. Please help triage it. Thank you!

PR #560, opened 2022-03-30
merge strings for train 30271

LZoog:
This is the result of the manual run of `extract-and-pull-request.sh` as requested.


locale/templates/auth.ftl
```diff
@@ -151,6 +150,37 @@ subscriptionUpdatePayment-plaintext = To prevent any interruption to your servic
...
+# Variables:
+#  $uaBrowser (String) - User's browser, e.g. Firefox
+#  $uaOS (String) - User's OS, e.g. Mac OSX
+device-browser-os = { $uaBrowser } on { $uaOS }
+
+# Variables:
+#  $ip (Number) - User's IP address
+user-ip = IP address: { $ip }
+
+# Variables:
```
peiying2, line 167: Do you want to use section comment like ## so it covers multiple strings?

locale/templates/auth.ftl
```diff
@@ -151,6 +150,37 @@ subscriptionUpdatePayment-plaintext = To prevent any interruption to your servic
...
+#  $city (String) - User's city
+#  $stateCode (String) - User's state
+#  $country (String) - User's country
+location-all = { $city }, { $stateCode }, { $country } (estimated)
```
peiying2, line 171: what does "estimated" mean here?

peiying2 (approved):

PR #574, opened 2022-04-29
Add new locales. Fix https://github.com/mozilla/fxa/issues/10292

clouserw:


clouserw:
@mathjazz does this look right to you?

peiying2 (approved):
Once Matjaz is ok with it, we can merge.

Issue #561, opened 2022-03-31
Odd markers in .ftl files (re: cancellationSurvey)


pdehaan:
Not sure if this is the best place for this issue, or how best to fix this. Came across it while trying to figure out a cheap HTML linter for .ftl files (which already caught a fair number of small issues).

Unfortunately, our template/reference string seems to look like this:

https://github.com/mozilla/fxa-content-server-l10n/blob/831e167cf2e0cb037614bb8d36c47822387cb29b/locale/templates/auth.ftl#L95

```diff
-<a data-l10n-name="cancellationSurveyUrl")s>short survey</a>
+<a data-l10n-name="cancellationSurveyUrl">short survey</a>
```

And I don't _think_ there should be an odd `)s` in our .ftl files. But that also means it was copied into about 44 translations as well.
If somebody can confirm that the `)s` is indeed unwanted, I should be able to quickly do suggestions in Pontoon, and we can fix the [locale/templates/auth.ftl:cancellationSurvey](https://github.com/mozilla/fxa-content-server-l10n/blob/831e167cf2e0cb037614bb8d36c47822387cb29b/locale/templates/auth.ftl#L95) string directly in GitHub (or however this machine works).

```text
git grep '")s>' locale/**/*.ftl

locale/cs/auth.ftl:cancellationSurvey = Pomozte nám vylepšit naše služby tím, že se zúčastníte tohoto <a data-l10n-name="cancellationSurveyUrl")s>krátkého průzkumu</a>.
locale/cy/auth.ftl:cancellationSurvey = Helpwch ni i wella ein gwasanaethau trwy lanw'r <a a data-l10n-name="cancellationSurveyUrl")s>arolwg byr</a> hwn.
locale/da/auth.ftl:cancellationSurvey = Hjælp os med at forbedre vores tjenester ved at deltage i denne <a data-l10n-name="cancellationSurveyUrl")s>korte undersøgelse</a>.
locale/de/auth.ftl:cancellationSurvey = Bitte helfen Sie uns, unsere Dienste zu verbessern, indem Sie an dieser <a data-l10n-name="cancellationSurveyUrl")s>kurzen Umfrage</a> teilnehmen.
locale/dsb/auth.ftl:cancellationSurvey = Pšosym wobźělśo se na toś tom <a data-l10n-name="cancellationSurveyUrl")s>krotkem napšašowanju</a>, aby nam pomagał, naše słužby pólěpšyś.
[...]

PR #578, opened 2022-05-06
Fix extra characters in data-l10n-name (cancellationSurvey)

flodolo:
Fixes issue #561 

This looks like an issue with the conversion from Django (`%(var)s`)  to Fluent.


flodolo:
I'm not completely sure how I managed to merge this by mistake (too many open tabs I guess), I meant to ask for review :-( 

@LZoog 
I think this is safe, but I also left a pull request in the main fxa to fix the source file
https://github.com/mozilla/fxa/pull/12788

PR #579, opened 2022-05-11
Use GH Actions to extract new strings

flodolo:
You can see a sample run on my fork here
https://github.com/flodolo/fxa-content-server-l10n/actions

And a resulting pull request
https://github.com/flodolo/fxa-content-server-l10n/pull/3

Note that I had to change the settings in my fork (Settings->Actions->General) to give permissions to write (create the branch) and open a pull request.

Schermata 2022-05-11 alle 09 34 48


.github/workflows/l10n_extract.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - cron: '0 22 * * 4' # At 15:00 PDT on Thursday
+  workflow_dispatch:
+jobs:
+  extract:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Linux packages
+        run: |
+          sudo apt update
+          sudo apt install gettext graphicsmagick git hub translate-toolkit -y
```
flodolo, line 13: Couple of notes here.

I'm using `hub`. The previous workflow was using the [gh npm package](https://www.npmjs.com/package/gh) which looks deprecated.

I'm also installing `translate-toolkit`. `merge_po.sh` has a conditional on `podebug` being available to generate `db_LB`, not sure if we want to skip this part.

.github/workflows/l10n_extract.yaml
```diff
@@ -0,0 +1,64 @@
...
+            --l10n-repo ../fxa-l10n
+      - name: Commit changes and open pull request
+        run: |
+          git config --global user.email 'bugmirror@restmail.net'
+          git config --global user.name 'Bug Mirror'
+          # Random release number to avoid collision with old trains or branches
+          TRAIN_NUMBER=$(( $RANDOM + $RANDOM + 1000 ))
+
+          cd fxa-l10n
+          git checkout -B "merge-train-$TRAIN_NUMBER-strings"
```
flodolo, line 56: I moved all the git commands out of the bash script, so no need to pass the random train number as parameter.

README.md
```diff
@@ -1,13 +1,13 @@
 fxa-content-server-l10n
 =======================
 
-[![CircleCI](https://circleci.com/gh/mozilla/fxa-content-server-l10n/tree/master.svg?style=svg)](https://circleci.com/gh/mozilla/fxa-content-server-l10n/tree/master)
+[![Tests](https://github.com/mozilla/fxa-content-server-l10n/actions/workflows/test.yaml/badge.svg)](https://github.com/mozilla/fxa-content-server-l10n/actions/workflows/test.yaml)
```
flodolo, line 4: This badge won't work until the action is run once. Here's the fork's badge https://github.com/flodolo/fxa-content-server-l10n/actions/workflows/test.yaml/badge.svg

README.md
```diff
@@ -63,12 +64,4 @@ This will create a new branch. Now run:
 ./scripts/before_prod_deploy.sh --l10n-repo .
 ```
 
-Git commit the changes made by the script. Pull request that branch against `master`.
-
-After that pull request is merged run:
-
-```
-./scripts/email-dev-l10n.sh
```
flodolo, line 71: We have decommissioned mailing list years ago, so this is not useful (also removed the script).

scripts/extract-and-pull-request.sh
```diff
@@ -1,14 +0,0 @@
-#!/bin/bash
```
flodolo, line 1: Moved this to the action. There's no need to run the script locally, as the action can be manually triggered as needed, without setting up local environment.

.github/workflows/l10n_extract.yaml
flodolo: Actually, it looks like this is failing because the locale doesn't exist. Given we're moving to Fluent, I guess we can ignore it (as we were doing in Circle CI)

```
podebug: WARNING: Error processing: input locale/templates/LC_MESSAGES/messages.pot, output locale/db_LB/LC_MESSAGES/messages.po, template None: [Errno 2] No such file or directory: 'locale/templates/LC_MESSAGES/messages.pot'
```

clouserw (approved):
I like it, this is a step in the right direction.  Let's land it this week before the cron time on Thursday so we can verify it works like we want.

.github/workflows/l10n_extract.yaml
flodolo: Confirmed, it's not needed.

PR #581, opened 2022-05-12
Make GetText extraction less noisy

flodolo:
The current process creates incredibly noisy pull requests (always more than 200 files), when the only thing changing is the POT creation date (which [changes](https://github.com/zaach/jsxgettext/blob/master/lib/jsxgettext.js#L144) every time). That's even worse, considering that these days we mostly only update Fluent files.

The approach is to create an hash of the .pot files without the POT creation date line before running extraction, and compare it after. If the hash changes, all the existing GetText commands are executed.

Also opted for moving stuff out of the deploy script, which AFAIK is only called within string extraction (and extract Sr-Latn twice).

Test locally that the Python script outputs the expected results when touching the .pot file, this is the result of running the updated script on my fork: https://github.com/flodolo/fxa-content-server-l10n/pull/6


scripts/extract_strings.sh
```diff
@@ -76,21 +76,53 @@ set -x
...
-cp -r $CONTENT_DIR/locale/templates/* $L10N_DIR/locale/templates
 
-# see https://github.com/mozilla/zippy/blob/180f746cd84cb3e409c76649e810485acb4512dc/locale/omg_new_l10n.sh#L90
-# also see https://bugzilla.mozilla.org/show_bug.cgi?id=1168488
-msgfilter -i $L10N_DIR/locale/sr/LC_MESSAGES/client.po -o $L10N_DIR/locale/sr_Latn/LC_MESSAGES/client.po recode-sr-latin
-msgfilter -i $L10N_DIR/locale/sr/LC_MESSAGES/server.po -o $L10N_DIR/locale/sr_Latn/LC_MESSAGES/server.po recode-sr-latin
+# Extract the hash of the pot files content without the POT-Creation-Date line.
+# This is used to avoid copying over the templates if the actual content
+# was not updated by the l10n-extract script
+OLD_HASHES=$(python $L10N_DIR/scripts/pot_checksum.py $L10N_DIR/locale/templates/)
```
flodolo, line 83: I'm sure this could have been done via Bash, but I didn't want to go down a rabbit hole of commands, and the Python script is pretty easy to go through, without requiring extra packages.

PR #585, opened 2022-05-17
Drop sr-Latn locale

### LINKED ISSUE #576: Drop sr-Latn?
That was a language that we used for Firefox OS, and it was generated automatically from Serbian (Cyrillic). 

Is there anything generating that now? I'm asking, because my linter caught an error in `verificationReminderFirst-description`(auth.ftl), and that was fixed a month ago in [Serbian](https://pontoon.mozilla.org/sr/firefox-accounts/auth.ftl/?search=erificationReminderFirst-description&string=235641).

flodolo:
Fixes #576 

I assume there's nothing to do in FxA? In Pontoon the locale doesn't exist.


clouserw:
We'd need to delete https://github.com/mozilla/fxa/blob/main/packages/fxa-shared/l10n/supportedLanguages.json#L64 and give a heads up to the sub plat team in case they are using it.  @biancadanforth - any concerns about dropping sr-LATN?

clouserw (approved):

clouserw:
I removed this from fxa in https://github.com/mozilla/fxa/issues/12950 and sub plat said to go ahead so this can land anytime.

PR #586, opened 2022-05-17
Include en and en-US FTL files in locale/

LZoog:
There's been some discussion lately on whether or not we _want_ to rely purely on fallback text for en/en-US users, as we have been since setting Fluent up in FxA as far as I know. We had a recent bug show us that English/US users were actually very likely seeing `en-GB` locales (see [more info here](https://github.com/mozilla/fxa/pull/12861#issuecomment-1125440236)), and since changing the Fluent strategy to be more precise in its matching, our fallback text is showing as expected. However, we keep uncovering instances where we haven't supplied fallback text, which has created several fairly severe (from a UI point of view) bugs, the newest few coming up today.

While discussions on how we handle fallback text are still ongoing and at the time of writing we want to always supply fallback text, the team seems to agree on at least on including our English FTL files in the `locale` directory for consistency, and I believe we should go ahead and do this now to prevent any further "fallback text missing" surprises when we push to production.

### DIFF
--- a/scripts/extract_strings.sh
+++ b/scripts/extract_strings.sh
@@ -102,11 +102,6 @@ else
     ./scripts/merge_po.sh ./locale
 fi
 
-# Fluent extraction
-cp $PAYMENTS_DIR/public/locales/en-US/*.ftl $L10N_DIR/locale/templates
-cp $SETTINGS_DIR/public/locales/en-US/*.ftl $L10N_DIR/locale/templates
-cp $MAILER_DIR/public/locales/en/*.ftl $L10N_DIR/locale/templates
-
 # Some locales must be copied over to a different locale code
 # See details in https://github.com/mozilla/fxa-content-server-l10n/pull/79#issuecomment-149281761
 cp -r $L10N_DIR/locale/sv_SE/* $L10N_DIR/locale/sv
@@ -124,5 +119,16 @@ sed -i'' -e 's/Language: pt_PT/Language: pt/g' "$L10N_DIR/locale/pt/LC_MESSAGES/
 cp -r $L10N_DIR/locale/fy_NL/* $L10N_DIR/locale/fy
 sed -i'' -e 's/Language: fy_NL/Language: fy/g' "$L10N_DIR/locale/fy/LC_MESSAGES/client.po"
 sed -i'' -e 's/Language: fy_NL/Language: fy/g' "$L10N_DIR/locale/fy/LC_MESSAGES/server.po"
+
+# Fluent extraction
+cp $PAYMENTS_DIR/public/locales/en-US/*.ftl $L10N_DIR/locale/templates
+cp $SETTINGS_DIR/public/locales/en-US/*.ftl $L10N_DIR/locale/templates
+cp $MAILER_DIR/public/locales/en/*.ftl $L10N_DIR/locale/templates
+
+# Pontoon will read from the "templates" directory but we must copy the FTL files
+# into "en" to ensure users with English will always see good fallback text.
+# We can remove the "templates" directory once we are off of gettext entirely -
+# then, Pontoon will read from the "en" directory.
+cp $L10N_DIR/locale/templates/*.ftl $L10N_DIR/locale/en
 
 set +x


flodolo (commented):
I'm not completely sure why we should be copying 2 files to `en-US` and 1 to `en`.

It feels like we should either copy all 3 files to both folders, or pick one, not split the content between the 2? If we chose one, `en` would seem like the best choice, but it depends on what the code does.

scripts/extract_strings.sh
```diff
@@ -103,9 +103,15 @@ else
...
 # Fluent extraction
+# fxa-payments-server
 cp $PAYMENTS_DIR/public/locales/en-US/*.ftl $L10N_DIR/locale/templates
+cp $PAYMENTS_DIR/public/locales/en-US/*.ftl $L10N_DIR/locale/en_US
+# fxa-settings
 cp $SETTINGS_DIR/public/locales/en-US/*.ftl $L10N_DIR/locale/templates
+cp $SETTINGS_DIR/public/locales/en-US/*.ftl $L10N_DIR/locale/en_US
+# fxa-auth-server
 cp $MAILER_DIR/public/locales/en/*.ftl $L10N_DIR/locale/templates
+cp $MAILER_DIR/public/locales/en/*.ftl $L10N_DIR/locale/en
```
flodolo, line 115: I would do the copy at the end once, with a comment explaining why we need to copy files once again.

```
# We need to copy the FTL files over to "en" to ensure users with English
# will always get good fallback content.
cp $L10N_DIR/locale/templates/*.flt $L10N_DIR/locale/en
```

But we also need to check with @mathjazz: this PR would results in pot files being in templates, as expected, but we'll have FTL files in `en`, `en-US`, and `templates`. Can you confirm this will not confuse Pontoon's logic?

flodolo:
Not very scientific, but looking at data for one random day (May 17), comparing FxA and Firefox Release English distribution.

| Locale | Firefox Accounts | Firefox |
| ----- | ----- | ----- |
| en-US | 82.67% | 92.1% |
| en-GB | 11.02% | 6.23% |
| en-CA | 2.52% | 1.70% |
| en | 3,80% | - |

This shows that en-GB is definitely over-represented.

scripts/extract_strings.sh
mathjazz: It will not confuse Pontoon. It takes source files from the `templates` folder. If it doesn't find them there, it checks `en-US` and then `en`.

I wonder if we can store all reference files in one directory (e.g. `en-US`) and take care of the fallbacks programmatically.

flodolo: > I wonder if we can store all reference files in one directory (e.g. en-US) and take care of the fallbacks programmatically.

I assume Pontoon only checks one folder out of the three (templates -> en-US -> en)? If that's correct, we need the .pot files in `templates` for GetText, but if we put the FTL files in `en-US` they won't be considered at all. So, we need all source files in `templates` for the time being (or until GetText is needed).

Could you clarify what you mean with the last part of the sentence?

mathjazz: Yes, Pontoon only checks one folder out of three (templates, en-US, en), which is why I'm wondering if that could be the case on the Firefox Accounts side as well.

From Pontoon's perspective there's no need for the `en-US` and `en` folders.

flodolo: Pontoon might not need them, but FxA needs a fallback locale with en-US content (that's the reason for this PR in the first place). 

`templates` would really not be an intuitive choice for that, as we always have either `en` or `en-US` in all existing Fluent projects.

LZoog:
> I'm not completely sure why we should be copying 2 files to en-US and 1 to en.

It feels like we should either copy all 3 files to both folders, or pick one, not split the content between the 2? If we chose one, en would seem like the best choice, but it depends on what the code does.


> I'm not completely sure why we should be copying 2 files to `en-US` and 1 to `en`.
> 
> It feels like we should either copy all 3 files to both folders, or pick one, not split the content between the 2? If we chose one, `en` would seem like the best choice, but it depends on what the code does.

Hmm yeah, I'm fine with copying them all into `en`, if you recommend that. It's on our todo list to be more consistent here, I split it out since in `fxa-payments-server` and `fxa-settings` we're calling all of our FTL files "en-US.ftl" and in `fxa-auth-server` we're calling them all `en.ftl` so it seemed copying them into those corresponding directories was most consistent based on the existing names.

flodolo:
Would it make sense to get rid of one between `en` and `en-US` in this repository? 

I would copy all the FTL files to `en`  and see if that solves the issue with fallback on your side, possibly drop `en_US` in the process.

scripts/extract_strings.sh
LZoog: > It will not confuse Pontoon. It takes source files from the templates folder. If it doesn't find them there, it checks `en-US` and then `en`.

Nice, I didn't realize it would check the `locale` directory if it couldn't find one from the `templates` directory. If I'm understanding this conversation correctly, it sounds like so long as we need gettext files or so long as gettext files are outputted to `templates`, that we can't remove our FTL files from `templates` in favor of relying on them being inside `locale/en` or `locale/en_US`? But then, when we do move off of gettext, we can potentially stop placing our FTL files into `templates`, so long as all of our FTL files are either in `locale/en` or `locale/en_US`. I don't think we want to alter any gettext files or output at the moment, we only have one package left to phase out before we're entirely on Fluent.

LZoog:
> Would it make sense to get rid of one between `en` and `en-US` in this repository?
> 
> I would copy all the FTL files to `en`  and see if that solves the issue with fallback on your side, possibly drop `en_US` in the process.

This sounds good to me, I'll update this PR shortly. You'll probably see another PR over here at some point in the future when we change our `en-US.ftl` files to just be `en.ftl` (payments & settings).

scripts/extract_strings.sh
LZoog: Thanks for clarifying. I pushed an update with a couple comments.

scripts/extract_strings.sh
```diff
@@ -125,4 +120,17 @@ cp -r $L10N_DIR/locale/fy_NL/* $L10N_DIR/locale/fy
...
+cp $SETTINGS_DIR/public/locales/en-US/*.ftl $L10N_DIR/locale/templates
+cp $MAILER_DIR/public/locales/en/*.ftl $L10N_DIR/locale/templates
+
+# Pontoon will read from the "templates" directory but we must copy the FTL files
+# into "en" to ensure users with English will always see good fallback text.
+# We can remove the "templates" directory once we are off of gettext entirely -
+# then, Pontoon will read from the "en" directory.
+cp $PAYMENTS_DIR/public/locales/en-US/*.ftl $L10N_DIR/locale/en
+cp $SETTINGS_DIR/public/locales/en-US/*.ftl $L10N_DIR/locale/en
+cp $MAILER_DIR/public/locales/en/*.ftl $L10N_DIR/locale/en
```
flodolo, line 134: Is there a reason not to use `cp $L10N_DIR/locale/templates/*.ftl $L10N_DIR/locale/en` vs 3 commands?

The advantage is that this line will remain the same, no matter if one of the packages change locale code later.

flodolo:
> I would copy all the FTL files to en  and see if that solves the issue with fallback on your side, possibly drop en_US in the process.

Can we safely remove en_US in this PR, or does that need changes in FxA?

scripts/extract_strings.sh
LZoog: I did that to see explicitly what's copied over but you're right, that is better. 😁 Pushed the update.

PR #604, opened 2022-06-02
Merge strings for train 45083

mozilla-pontoon:


locale/en/settings.ftl
```diff
@@ -115,10 +115,9 @@ cs-disconnect-sync-heading = Disconnect from Sync
 # Variables:
 #   $device (String) - the name of a device using Firefox Accounts
 #                      (for example: "Firefox Nightly on Google Pixel 4a")
-cs-disconnect-sync-content = Your browsing data will remain on your
-  device ({ $device }), but it will no longer sync with your account.
-cs-disconnect-sync-reason = What’s the main reason for disconnecting this
-  device?
+cs-disconnect-sync-content-2 = Your browsing data will remain on { $device },
```
flodolo, line 118: @bcolsson 
In these PRs we only need to look at either `en` or `templates`, since they're the same. `fy` is just a by-product of the extraction, copying translations from `fy-NL`, so it can be safely ignored (same for `es`, `hi`, `sv`).

This PR is changing two existing strings, and using a new ID for that, which is what we require per [documentation](https://mozilla-l10n.github.io/documentation/localization/making_string_changes.html).

But it's also introducing a small issue: now the second string also has a variable `$device`, but the comment is still a "single string" comment. So, in Pontoon it won't show up for the second string, only for the first. The correct fix should have been to change that to a [group comment](https://projectfluent.org/fluent/guide/comments.html) (two hashes, one empty line after).

Let me know if everything is clear. If so, feel free to approve the pull request, and hit the _Squash and merge_ button, so we can test your rights ;-) 

Unfortunately, we can fix it here, and that's why we set up review directly in the `FxA` project.

bcolsson: Okay thanks. Spent a bit of time going through step by step everything that lead to the PR (workflow, scripts, fxa repository) for repetition to improve my understanding of the entire flow. So at this point even if we changed it here, changes would get reverted the next time strings are extracted (or perhaps it might even make issues for pontoon if we changed it here)? And the correct fix would be to make the changes to the [source document](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/components/ConnectedServices/en-US.ftl)?

flodolo: > So at this point even if we changed it here, changes would get reverted the next time strings are extracted (or perhaps it might even make issues for pontoon if we changed it here)?

That's correct. We could fix it here, but it risks getting overwritten the next time we export strings. No risks on Pontoon side though, especially considering it's only a comment. I would definitely not fix a string, for example.

> And the correct fix would be to make the changes to the [source document](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/components/ConnectedServices/en-US.ftl)?

Again, correct :-) 

With the new CODEOWNERS hook there, we would be flagged as reviewer in the pull request, and we would be able to catch issues before they land.

Talking from experience, especially if the volume gets high, things might still slip through. The good thing is that you have another step in this phase (fxa -> fxa-l10n) to catch them.

Congrats on your first PR merge 🥳

bcolsson: Wooh, thanks! 🥳

PR #615, opened 2022-07-13
Commiting github action to evaluate fluent files for change strings

bcolsson:
wthout associated ids


bcolsson:
@flodolo - low priority, but created an action that parses fluent files. I borrowed from the fluent-python parser and did a bit of testing but would appreciate a review when you are back from PTO.

.github/scripts/unchanged-ids/fluent.syntax/parse_messages_terms_to_dict.py
```diff
@@ -0,0 +1,67 @@
...
+        files = list(errors.keys())
+        files.sort()
+
+        output = []
+        total_errors = 0
+        for file in files:
+            output.append(f"\nFile: {file}")
+            for id in errors[file]:
+                output.append(
+                    f"\nID: {id[0]}"
```
flodolo, line 53: It would be more readable to have a dictionary with meaningful key names than relying on list indexes.

flodolo (changes requested):
While this mostly works, I think there are a few issues to fix.

The main problem is copying over code from an existing package, which makes maintenance a problem. Technically you could extend the existing class, but for parsing strings you should rely directly on compare-locales without going into the details of fluent-syntax.

This [snippet](https://github.com/mozilla-l10n/mozl10n-linter/blob/main/l10n/fluent_l10n.py#L87-L122) should give you a good starting point.

Other notes:
* I wouldn't rely on relative paths. You can extract the path of the script to determine the root of the folder, then use absolute paths. For example (assuming the script is placed in `.github/scripts`:
```
    script_path = os.path.dirname(__file__)
    root_path = os.path.abspath(os.path.join(script_path, os.pardir, os.pardir))
 ```
* This would fail for the pull requests updating fake locales (e.g. if Spanish updates an existing translation). We should only check files for reference locale, and that info should be passed as an argument to the script for portability.

flodolo:
Example of extending the Serializer class and changing some function, without copying over the whole package (should be functionally equivalent)
https://www.diffchecker.com/XklKQLzE

But, as said, we should just parse with compare-locales ;-)

flodolo:
One more thought: this should probably be a step in the [existing workflow](https://github.com/mozilla/fxa-content-server-l10n/blob/master/.github/workflows/reference_linter.yaml) checking reference files.


> * This would fail for the pull requests updating fake locales

I also realize this is wrong, because it would trigger for any commit for a locale updating existing translations.

PR #623, opened 2022-07-21
Merge strings for train 25435

mozilla-pontoon:


locale/templates/LC_MESSAGES/client.pot
```diff
@@ -2110,7 +2124,7 @@ msgid "How old are you?"
 msgstr ""
 
 #: .es5/templates/partial/coppa-age-input.mustache:3
-msgid "You need to be at least 13 years old to create an account."
+msgid "You must be over 13 to create an account."
```
bcolsson, line 2127: @millsoper It appears the meaning of this string is being changed and introducing an error. Please confirm this wording is correct. 
"at least 13 years old" is inclusive of 13, while "over 13" is exclusive. My understanding is that children under the age of 13 (so ages 0 to 12) are subject to COPPA, while 13 and older are not.

PR #625, opened 2022-07-23
Linter: check for string changes without new IDs

bcolsson:
Revised PR based on feedback from #615
- Now uses compare-locales to parse
- Bash script no longer changes filenames, instead moves files to old/ and new/ folders
- Integrated into reference locale linter action

### DIFF
--- a/.github/scripts/detect_unchanged_ids.py
+++ b/.github/scripts/detect_unchanged_ids.py
@@ -1,0 +1,113 @@
+import os
+import sys
+import argparse
+
+try:
+    from compare_locales import parser
+except ImportError as e:
+    print("FATAL: make sure that dependencies are installed")
+    print(e)
+    sys.exit(1)
+
+
+def get_file_list(path):
+    """Get the list of supported files."""
+    file_list = []
+    for root, dirs, files in os.walk(path, followlinks=True):
+        for file in files:
+            if file.endswith(".ftl"):
+                file_list.append(os.path.join(root, file))
+    file_list.sort()
+
+    return file_list
+
+
+def generate_dict(file_name):
+    file_parser = parser.getParser(".ftl")
+    file_parser.readFile(file_name)
+    try:
+        entities = file_parser.parse()
+        entity_dict = {}
+        for entity in entities:
+            # Ignore Junk
+            if isinstance(entity, parser.Junk):
+                continue
+            string_id = f"{entity}"
+            if entity.raw_val is not None:
+                entity_dict[string_id] = entity.raw_val
+            # Store attributes
+            for attribute in entity.attributes:
+                attr_string_id = f"{entity}.{attribute}"
+                entity_dict[attr_string_id] = attribute.raw_val
+        return entity_dict
+    except Exception as e:
+        print(f"Error parsing file: {file_name}")
+        print(e)
+
+
+def find_changed_ids(file_old, file_new):
+    dict_old = generate_dict(file_old)
+    dict_new = generate_dict(file_new)
+    return {
+        key: {"previous": dict_old[key], "new": dict_new[key]}
+        for key in dict_old.keys()
+        if key in dict_new and dict_old[key] != dict_new[key]
+    }
+
+
+def main():
+    arg_parser = argparse.ArgumentParser()
+    arg_parser.add_argument(
+        "--base_dir",
+        required=True,
+        help="Folder with the previous version of files (copied from base repository)",
+    )
+    arg_parser.add_argument(
+        "--head_dir",
+        required=True,
+        help="Folder with the new version of files (copied from head repository)",
+    )
+    args = arg_parser.parse_args()
+
+    errors = {}
+
+    script_path = os.path.dirname(__file__)
+    root_path = os.path.abspath(os.path.join(script_path, os.pardir, os.pardir))
+
+    path_old = os.path.join(root_path, args.base_dir)
+    path_new = os.path.join(root_path, args.head_dir)
+
+    files_old = get_file_list(path_old)
+    files_new = get_file_list(path_new)
+
+    for file_old, file_new in zip(files_old, files_new):
+        filename = os.path.relpath(file_old, path_old)
+        errors[filename] = find_changed_ids(file_old, file_new)
+
+    if errors:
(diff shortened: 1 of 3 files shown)


.github/scripts/detect_unchanged_ids.py
```diff
@@ -0,0 +1,107 @@
...
+    return {
+        key: {'before': dict_old[key], 'after': dict_new[key]}
+        for key in dict_old.keys()
+        if key in dict_new and dict_old[key] != dict_new[key]
+    }
+
+
+def main():
+    arg_parser = argparse.ArgumentParser()
+    arg_parser.add_argument("dir_old")
```
flodolo, line 60: Please make these named arguments and add some description, at least it's impossible to swap the files by mistake, e.g.

```
    parser.add_argument(
        "--old",
        required=True,
        help="Folder with the previous version of the file",
    )
```

.github/scripts/detect_unchanged_ids.py
```diff
@@ -0,0 +1,107 @@
...
+
+    path_old = os.path.join(root_path, args.dir_old)
+    path_new = os.path.join(root_path, args.dir_new)
+
+    files_old = get_file_list(path_old)
+    files_new = get_file_list(path_new)
+
+
+    for file_old, file_new in zip(files_old, files_new):
+        filename = file_old.replace(f"{path_old}{os.sep}", "")
```
flodolo, line 77: Unless I'm misunderstanding the intent here:

```suggestion
        filename = os.path.relpath(file_old, path_old)
```

.github/scripts/detect_unchanged_ids.py
```diff
@@ -0,0 +1,107 @@
...
+        total_errors = 0
+        for file in files:
+            output.append(f"\nFile: {file}")
+
+            ids = list(errors[file].keys())
+            ids.sort()
+            for id in ids:
+                output.append(
+                    f"\nID: {id}" 
+                    f"\nBefore: {errors[file][id]['before']}"
```
flodolo, line 94: Maybe Previous/New instead of Before/After?

.github/scripts/detect_unchanged_ids.py
```diff
@@ -0,0 +1,107 @@
...
+                output.append(
+                    f"\nID: {id}" 
+                    f"\nBefore: {errors[file][id]['before']}" 
+                    f"\nAfter: {errors[file][id]['after']}"
+                )
+                total_errors += 1
+        output.append(f"\nTotal errors: {total_errors}")
+
+        print("\n".join(output))
+        sys.exit(1)
```
flodolo, line 101: This should work without a separate print.

```suggestion
        sys.exit("\n".join(output))
```

flodolo (changes requested):
This looks good, although I haven't tested it yet.

While this is triggered only for changes to `locale/en`, it will still analyze (and fail) for other locales. We need to limit checks to the reference locale by adding an explicit argument.

flodolo:
Forgot one note: don't forget to run `black` on the script.

flodolo:
Couple of things I forgot:
* Can you rebase against the current default branch? There were a few changes in workflows
* Run black on the script (I could spot a double empty line)

.github/scripts/detect_unchanged_ids.py
bcolsson: I was looking for something in the path library that would do it, but didn't quite figure out that relpath would do it. Thanks!

.github/scripts/detect_unchanged_ids.py
bcolsson: Would we not need an exit code to indicate the failure?

.github/scripts/detect_unchanged_ids.py
bcolsson: Fixed

.github/scripts/detect_unchanged_ids.py
flodolo: Anything but zero would do, so existing with the string has the same effect.

bcolsson:
I messed up the rebase pretty badly, so diff isn't saved, but made the following changes:
- Renamed folders/ updated python script call in workflow
- Implemented all your feedback

As I write this I remember I forgot to fix the issue where we need to restrict to only reference files, so that change is incoming.

.github/scripts/detect_unchanged_ids.py
```diff
@@ -0,0 +1,116 @@
...
+
+try:
+    from compare_locales import parser
+except ImportError as e:
+    print("FATAL: make sure that dependencies are installed")
+    print(e)
+    sys.exit(1)
+
+
+def get_file_list(path, reference_locale="locale/en"):
```
flodolo, line 13: The goal would be to be able to reuse this in other places, so it needs to be a parameter that you pass to the script, and also an actual locale code (`en` in this case)

.github/workflows/reference_linter.yaml
```diff
@@ -30,3 +32,13 @@ jobs:
       - name: Lint reference
         run: |
           moz-fluent-lint ./locale/en --config .github/linter_config.yml
+      - name: Check for unchanged IDs
+        run: |
+            diff=$( git diff --name-only ${{ github.event.pull_request.base.sha }} $GITHUB_SHA )
+            for file in $diff; do
+              mkdir -p "old/${file%/*}/" && mkdir -p "new/${file%/*}/"
```
flodolo, line 39: You missed updating this line, so the workflow is currently broken.

flodolo (changes requested):
You lost the change to requirements for compare-locales, so the workflow is currently not running. 

I suggest to set it up on master in your fork for testing, then you can create pull requests against your own fork to make sure it's working.

Heads-up: always double check that you're making PRs against your fork, the default will be to use this repo, and I've made that mistake before (as recently as last week…)

.github/scripts/detect_unchanged_ids.py
flodolo: On second thoughts, I see how using a locale code might make things harder, so we can pass the "path to reference locale" as argument.

flodolo (approved):
Looks good now 👍🏼

PR #637, opened 2022-08-20
Test removing obsolete strings

flodolo:


flodolo:
Need to set maxwidth and delete .tmp files

PR #640, opened 2022-08-22
Update string across locales to include 'br' class

LZoog:
In [this pull request](https://github.com/mozilla/fxa/pull/13967), we needed to add the `mb-3` class to this `br`. To help over here, I found-and-replaced the existing English string and added the class to the translated strings.

Couple questions:
1) Is this okay even if the line number changes? In our code, the line changed by 2, but I'm not sure what the output will look like until we build / if I were to run the extract script. Not sure if this matters.
2) Is this PR helpful, or would this kind of change not matter much to Pontoon/translators because the string in Pontoon's memory is almost identical? Do we have docs on how Pontoon works with strings? I have a couple of notes but, I've been meaning to update FxA's l10n docs and might have some similar questions.

### DIFF
--- a/locale/ar/LC_MESSAGES/client.po
+++ b/locale/ar/LC_MESSAGES/client.po
@@ -1500,7 +1500,7 @@ msgid "Repeat password"
 msgstr "كرر كلمة السرّ"
 
 #: .es5/templates/sign_up_password.mustache:6
-msgid "You need this password to access any encrypted data you store with us.<br/>A reset means potentially losing data like passwords and bookmarks."
+msgid "You need this password to access any encrypted data you store with us.<br class=\"mb-3\">A reset means potentially losing data like passwords and bookmarks."
 msgstr ""
 
 #: .es5/templates/sign_up_password.mustache:7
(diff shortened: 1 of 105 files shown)


flodolo:
> 1. Is this okay even if the line number changes? In our code, the line changed by 2, but I'm not sure what the output will look like until we build / if I were to run the extract script. Not sure if this matters.

What do you mean "changed by 2"?

> 2. Is this PR helpful, or would this kind of change not matter much to Pontoon/translators because the string in Pontoon's memory is almost identical?

The challenge is that such a PR is going to get obsolete quickly, if that file is being actively translated (also because the output of msgmerge and Pontoon is slightly different). This type of PR should also update the templates files.

In general, it's safer to just change the string in the code repository, and have it retranslated via Pontoon. Pontoon should show that there's a very similar string already translated, with a diff of what changed.

This is an example for recent similar changes done in FxA .po files

Schermata 2022-08-22 alle 18 48 26

I'm going to close this PR, and just let the string go through the standard process. But thanks for trying to help out localizers 👍🏼

PR #644, opened 2022-09-07
Merge strings for train 50800

mozilla-pontoon:


bcolsson:
Ran this to test if extract working. Some strings might be fixed before tomorrow so closing this to wait for the scheduled extract.

PR #654, opened 2022-10-26
Simplify fxa-settings extract strings command

LZoog:
With [this PR and change]( https://github.com/mozilla/fxa/pull/14312/files#diff-7476db00402b40902a9cf3a6834136ed6356ae6591cb4c40a2ed99c2c0a5464fR22), we're using grunt instead of webpack to create our `settings.ftl` file. Webpack required a full build to concat our files, and while `build` still works, since it's now separate from webpack we can run the grunt task instead to make this workflow slightly faster.

### DIFF
--- a/.github/workflows/l10n_extract.yaml
+++ b/.github/workflows/l10n_extract.yaml
@@ -36,7 +36,7 @@ jobs:
         run: |
           cd fxa-code
           yarn workspaces focus fxa-content-server fxa-auth-server fxa-payments-server fxa-settings
-          yarn workspace fxa-settings build
+          yarn workspace fxa-settings grunt merge-ftl
           yarn workspace fxa-auth-server grunt merge-ftl
 
           NODE_ENV=development ../fxa-l10n/scripts/extract_strings.sh \


.github/workflows/l10n_extract.yaml
```diff
@@ -36,7 +36,7 @@ jobs:
         run: |
           cd fxa-code
           yarn workspaces focus fxa-content-server fxa-auth-server fxa-payments-server fxa-settings
-          yarn workspace fxa-settings build
+          yarn workspace fxa-settings grunt merge-ftl
```
bcolsson, line 39: @LZoog 

So from local testing, it appears the concat file appears in the `en-US` locale when built with `grunt merge-ftl`, but currently fxa-content-server-l10n uses the en locale for fxa-settings. I'm having trouble getting `yarn workspace fxa-settings build` to work locally so I can't confirm my hypothesis but my assumption is that fxa-settings used to only generate the `en` locale. 

I think this [line](https://github.com/mozilla/fxa/blob/befccf53df44eebe3422a384a8a5b8930f8bcad8/packages/fxa-settings/Gruntfile.js#L14) might need to be adjusted to output to en. 

I confirmed that fxa-auth-server builds into the `en` locale.

flodolo: Looking at the script in the l10n repository, I think we expect the file to be generated in en-US?
https://github.com/mozilla/fxa-content-server-l10n/blob/main/scripts/extract_strings.sh#L121

bcolsson: 😓 
There’s the step I missed. Apologies, ignore me.

LZoog: We do currently expect it to be generated in `en-US`. We have [FXA-6003](https://mozilla-hub.atlassian.net/browse/FXA-6003) filed and reference that line^ in this repo to change in that ticket. 🙂 I appreciate the thoroughness.

bcolsson (approved):

PR #662, opened 2022-11-10
Merge strings for train 45075

mozilla-pontoon:


bcolsson (approved):

mathjazz:
We need this patch to also:
- rename all locale `main.ftl` files to `payments.ftl`
- remove the `main.ftl` file from the source folder(s)

I'm curious:
- what is the purpose of having en, en-US and templates folders?
- why are some l10n files getting updates in this PR?

mathjazz:
> templates is needed for gettext. Once we remove that, I think we can live with only one locale folder, but we need to confirm with Matjaz that Pontoon will pick the right folder as source at that point.

Pontoon is fine with any of these as a source folder. Seems like we could already remove `en` and `en-US`? Anyways, not relevant for this PR, I just noticed that we have three and they aren't in sync.

bcolsson:
Closing this PR due to conflict, new PR is #663

PR #665, opened 2022-11-24
Merge strings for train 11032

mozilla-pontoon:


flodolo (approved):

flodolo:
Actually holding on these strings, since it feels like a waste of resource to have to translate shared strings twice.

PR #667, opened 2022-11-30
Extract strings from fxa-react, fxa-shared

vpomerleau:
PR linked with [Fxa PR-14567](https://github.com/mozilla/fxa/pull/14567) - awaiting review, these two PRs should only be merged when both approved

In response to [FXA-14548 on GitHub](https://github.com/mozilla/fxa/issues/14548)/[#FXA-6388 on Jira](https://mozilla-hub.atlassian.net/browse/FXA-6388)

Because:

- Strings from shared components and branding that were concatenated into multiple packages before string extraction were causing duplication on Pontoon.

This commit:

- Add Fluent String extraction for two more dir (`fxa-react` and `fxa-shared`)
- Extract strings from specific files (e.g., change `*.ftl` to `payments.ftl`)

### DIFF
--- a/scripts/extract_strings.sh
+++ b/scripts/extract_strings.sh
@@ -28,6 +28,8 @@ MAILER_DIR="./fxa-auth-server"
 CONTENT_DIR="./fxa-content-server"
 PAYMENTS_DIR="./fxa-payments-server"
 SETTINGS_DIR="./fxa-settings"
+REACT_DIR="./fxa-react"
+SHARED_DIR="./fxa-shared"
 L10N_DIR="./fxa-content-server-l10n"
 
 while [[ $# > 1 ]]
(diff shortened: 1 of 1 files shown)


bcolsson (approved):
This seems pretty straight forward. Looks good (though the shared.ftl being in react and not Shared DIR is a bit confusing, but I suppose can't be helped.)

vpomerleau:
> the shared.ftl being in react and not Shared DIR is a bit confusing

I agree, I had the same thought. We'd floated the idea of `react-shared`, which seemed awkward, but perhaps we could just go with `react.ftl` for clarity? @dschom do you have thoughts on naming?

vpomerleau:
@bcolsson @dschom  I've renamed shared.ftl to react.ftl. If there are any objections, please let me know!

bcolsson:
Looks good to me. Holding until the other patch gets merged on FXA.

PR #681, opened 2023-01-13
Reading env vars for circle ci incident

jbuck:


flodolo:
@jbuck 
Sorry, I don't understand. We have stopped using CircleCI months ago ([in May](https://github.com/mozilla/fxa-content-server-l10n/pull/579)) for this repository, why this change?

jbuck:
I needed to read the encrypted values https://github.com/mozilla/fxa-content-server-l10n/blob/bbdc20efdf3dadc8f4bf758b5ead223d6f9e2bf7/id_bug_mirror.enc and https://github.com/mozilla/fxa-content-server-l10n/blob/bbdc20efdf3dadc8f4bf758b5ead223d6f9e2bf7/secret-env-cipher . The only way to get the encryption key was to run a Circle CI job with SSH enabled. I've reverted this PR in https://github.com/mozilla/fxa-content-server-l10n/commit/8f2d040aa301cd61f1b16d2926b12cc9d31dfa5c now that I have the encryption key

PR #704, opened 2023-03-20
Merge strings for train 39598

mozilla-pontoon:


bcolsson:
Confirming string IDs don't need updates for these changes

PR #707, opened 2023-03-24
Update extract workflow run cron schedule

bcolsson:


.github/workflows/l10n_extract.yaml
```diff
@@ -1,7 +1,8 @@
 name: Extract Strings
 on:
   schedule:
-    - cron: '0 22 * * 4' # At 15:00 PDT on Thursday
+    - cron: '0 20 * * 1' # At 13:00 PDT on Mondays
+    - cron: '0 20 * * 4' # At 13:00 PDT on Thursdays
```
flodolo, line 5: Does this work? `0 20 * * 1,4` definitely would
https://crontab.guru/#0_20_*_*_1,4

flodolo: Well, it would work, but that feels like a very special case
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

bcolsson: Oops, I just skimmed that page.. and didn't realize the example they were showing was a special case. I'll update, thanks for pointing this out.

bcolsson: Revised in #708

PR #713, opened 2023-04-10
Merge strings for train 12441

mozilla-pontoon:


flodolo (approved):
Merging, since it's only sync fake locales.

PR #741, opened 2023-07-13
Merge strings for train 25863

mozilla-pontoon:


locale/templates/payments.ftl
```diff
@@ -31,6 +31,47 @@ project-brand = Firefox Accounts
...
+-product-firefox-monitor = Firefox Monitor
+-product-firefox-relay = Firefox Relay
+-product-pocket = Pocket
+
+-brand-paypal = PayPal
+-app-store = App Store
+-brand-google = Google
+-google-play = Google Play
+-brand-apple = Apple
+
```
bcolsson, line 74: The entire branding.ftl is getting imported into payments.ftl, causing duplication of brand terms.

PR #745, opened 2023-07-20
Merge strings for train 47920

mozilla-pontoon:


bcolsson (approved):
Failing check ok (no need for new id for a change in capitalization)

PR #855, opened 2024-09-27
Merge strings for train 21220

mozilla-pontoon:


bcolsson:
Fix was still pending after merge. Need to run again after it finishes.

PR #884, opened 2025-01-07
feat(next): add support for payments-next.ftl

StaberindeZA:
Because:

- The Subscription Platform team is adding a new app service, payments-next, to replace the existing fxa-payments-server. In the short term both payments-next and fxa-payments-server need to be supported, each utilizing different ftl strings.

This commit:

- Adds support for new app payments-next utilizing payments-next.ftl

Closes #FXA-7482

### DIFF
--- a/.github/workflows/l10n_extract.yaml
+++ b/.github/workflows/l10n_extract.yaml
@@ -41,10 +41,12 @@ jobs:
           yarn workspace fxa-settings grunt merge-ftl
           yarn workspace fxa-auth-server grunt merge-ftl
           yarn workspace fxa-react grunt merge-ftl
+          npx grunt --gruntfile='apps/payments/next/Gruntfile.js' merge-ftl
 
           NODE_ENV=development ../fxa-l10n/scripts/extract_strings.sh \
             --mailer-repo ./packages/fxa-auth-server \
             --payments-repo ./packages/fxa-payments-server \
+            --payments-next-repo ./apps/payments/next \
             --content-repo ./packages/fxa-content-server \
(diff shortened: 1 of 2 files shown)


StaberindeZA:
@bcolsson as discussed last year, this PR will add support for the new Subscription Platform app service Payments Next, as part of the Subscription Platform 3.0 project.

Along with this change, there is also a script that has been created that ports over existing translations from `locale/**/payments.ftl` to `locale/**/payments-next.ftl` where an existing translation is available.

I'll reach out to you directly to discuss further.

.github/workflows/l10n_extract.yaml
```diff
@@ -41,10 +41,12 @@ jobs:
           yarn workspace fxa-settings grunt merge-ftl
           yarn workspace fxa-auth-server grunt merge-ftl
           yarn workspace fxa-react grunt merge-ftl
+          yarn grunt --gruntfile='apps/payments/next/Gruntfile.js' merge-ftl
```
bcolsson, line 44: Testing this locally, the script fails when I attempt to run the command locally in terminal with `yarn grunt --gruntfile='apps/payments/next/Gruntfile.js' merge-ftl` unless I install things via `yarn install` with the error: `Usage Error: Couldn't find a script named "grunt".`

This is also failed when I attempted to run it from the fork: https://github.com/bcolsson/fxa-content-server-l10n/actions/runs/12658750573/job/35276377203

bcolsson (changes requested):

bcolsson (approved):

PR #914, opened 2025-03-10
Merge strings for train 34297

mozilla-pontoon:


bcolsson:
Following up with FxA team as there's a number of issues with some strings.

PR #915, opened 2025-03-13
Merge strings for train 62062

mozilla-pontoon:


bcolsson:
Removed strings from file that need fixing.
Linter issue raising unchanged ID is ok as its a style change only.

bcolsson (approved):

PR #1079, opened 2026-08-26
Update CI dependencies from dependabot, update node version to match FxA

bcolsson:
Tested locally and manually triggering automation against this branch and everything appears to work properly.

### DIFF
--- a/.github/workflows/reference_linter.yaml
+++ b/.github/workflows/reference_linter.yaml
@@ -36,11 +36,11 @@ jobs:
           moz-fluent-lint ./locale/en --config .github/linter_config.yml
       - name: Check for unchanged IDs
         run: |
-            diff=$( git diff --name-only ${{ github.event.pull_request.base.sha }} $GITHUB_SHA )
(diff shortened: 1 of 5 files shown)


flodolo (changes requested):
None of the checks ran. We should modify the workflows so that a change to the file itself trigger one.

flodolo:
I made several changes, so this needs some review from your side.

Once this is merged, we should rebase the two dependabot PRs and see if they still work.

.github/workflows/reference_linter.yaml
```diff
@@ -34,11 +36,11 @@ jobs:
           moz-fluent-lint ./locale/en --config .github/linter_config.yml
       - name: Check for unchanged IDs
         run: |
-            diff=$( git diff --name-only ${{ github.event.pull_request.base.sha }} $GITHUB_SHA )
+            diff=$( git diff --name-only --diff-filter=M ${{ github.event.pull_request.base.sha }} $GITHUB_SHA -- 'locale/en/*.ftl' )
```
bcolsson:
```suggestion
            diff=$( git diff --name-only --diff-filter=M ${{ github.event.pull_request.base.sha }} $GITHUB_SHA -- 'locale/templates/*.ftl' )
```
templates is the source locale.

